PR #21, opened 2021-05-16
feat: Plugin API.

### LINKED ISSUE #20: Neogit integration
Hello!

I really like the look of your plugin so I would like to integrate your plugin with [neogit](https://github.com/TimUntersberger/neogit).

Some things are still a bit unclear for me so I'd appreciate it if you could help me with a few things:

* Is there a way to get notified when something happens? 
  * for example you commit a change; how can I get notified?
* Are you going to provide an API which can be used by other plugins? 
  * I am mainly looking for ensurance that the API won't have any breaking changes too often
* Could you support feeding the plugin the files myself? Right now opening the diff is a bit too slow for my taste and neogit already has the information when trying to open the diff.
* Is any merge conflict support planned or even wanted?

sindrets:
Fixes #20.

@TimUntersberger please take a look at this. Let me know if there are any more hooks/callbacks/methods you need exposed.

[Here's a example of how to use the API.](https://gist.github.com/sindrets/b750723f5f23182904f70a6274106304)

### DIFF
--- a/doc/diffview.txt
+++ b/doc/diffview.txt
@@ -147,5 +147,75 @@ R                       Update the stats and entries in the file list.
 
 <leader>e               Bring focus to the file panel.
 
+API                                              *diffview-api*
+
+Diffview provides an API that lets plugins supply custom file data to create
+custom diffviews.
+
+Example usage:
+https://gist.github.com/sindrets/b750723f5f23182904f70a6274106304
+
+CView                                             *diffview.api.c-view.CView*
+
+    Class that represents a custom Diffview.
+
+    CView:new({opt})
+
+        Parameters: ~
+            {opt}   (table)   Table containing the args passed to the view
+                              constructor.
+
+        Fields: ~
+            {git_root} (string)         Absolute path to the toplevel git
+                                        directory.
+            {left} (Rev)                The git rev representing the left
+                                        window in the diffview.
+            {right} (Rev)               The git rev representing the right
+                                        window in the diffview.
+            {files} (FileData[])        List of file data used to create the
+                                        file entries in the view.
+            {update_files} (function)   This function should return an updated
+                                        list of `FileData`, and is called from
+                                        `View:update_files`.
+                Parameters: ~
+                    {view} (CView)  The CView object that called the function.
+
+            {get_file_data} (function)  This should return a list of lines that
+                                        make up the buffer indicated by `path`
+                                        and `split`.
+                Parameters: ~
+                    {path} (string)           Path to the file, relative to
+                                              git root.
+                    {split} ("left"|"right")
+
+FileData                                         *diffview.api.c-view.FileData*
+
+    Table that contains the metadata used to create file entries.
+
+    Fields: ~
+        {path} (string)         Path relative to git root.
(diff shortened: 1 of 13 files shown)


TimUntersberger:
This looks good for now, but I can't really get any performance improvements by doing it myself.

I looked at a custom solution I implemented before and saw that I fetched the file content lazy instead of ahead of time.

I also need to be able to set the focused file. Right now I am doing this:

```lua
  local view

  await(util.time_async("OPENING DIFFVIEW", async(function()
    view = await(M.open())
  end)))

  util.time("SETTING DIFFVIEW FILE", function()
    for i=1,#view.files do
      local file = view.files[i]
      if file.path == file_name then
        view:set_file(file, true)
        break
      end
    end
  end)

  return view

```

`OPENING DIFFVIEW` usually takes around 400-500ms
`SETTING DIFFVIEW FILE` takes around 70-80ms

For context I got these numbers with `6` files in the diff.

If we somehow can make the most of the stuff lazy and somehow support setting the inital file the performance would improve a lot.

My solution API-wise would be to just have me provide two functions for navigating through the files, where I could then load the content JIT.

```lua
local view = CView:new {
  git_root = git_root,
  left = left,
  right = right,
  current_file = file,
  next_file = function()
    ...
  end,
  prev_file = function()
    ...
  end
}
```

Maybe there are other problems too, but I can at least asure you that in my custom solution this gives me very good performance (almost instantaneous)

sindrets:
The normal diffview does load diffs lazily, so I agree, we should implement that for the API as well. I'll revise the API for file loading.

> `OPENING DIFFVIEW` usually takes around 400-500ms
> `SETTING DIFFVIEW FILE` takes around 70-80ms

That seems strange to me. I have never seen runtimes that slow, even with views with far more than 6 files. Could you provide an example git repo with an example git rev that I can use to test?

Providing callbacks for `next_file` and `prev_file` is not the way to do it in my opinion, because you should be able to pick any file from the file panel. In which case you are not necessarily cycling through the files in order. 

> I also need to be able to set the focused file.

You want me to add a method to set the focused file by path? It would basically just be doing what you are already doing.

> and somehow support setting the inital file

Yep, good idea.

sindrets:
@TimUntersberger alright I changed the API. You now provide a function:

```lua
---This should return a list of lines that make up the buffer indicated by
---`path` and `split`.
---@param path string Path relative to git root.
---@param split "left"|"right"
---@return string[]|nil
function get_file_data(path, split)
  -- ...
end
```

This way the data can be provided lazily on  demand.

I added a method `View:set_file_by_path` that - as you might have guessed - lets you set the selected file by path (relative to git root).

I also added a new property to `FileData`: `selected`, that if set to true indicates that that file should be the initially selected file.

I updated the gist to reflect these changes.

TimUntersberger:
Thank you!

I implemented the new version according to your changes and displaying the diffview window now only takes a few milliseconds!

I also looked a bit at how merging works with the builtin diff and it looks like this shouldn't be too hard to add. 
I'd like to discuss this in its own issue though.

This is ready to merge on my side. 👍

PR #36, opened 2021-07-03
Use <Cmd> instead of :

Shatur:
When executing commands in scripts or hotkeys, it is always better to use `<Cmd>` instead of `:`. This is because the colon raises cmdline opening events. For example, in my UI it causes weird animation.
Also `<Cmd>` is faster :)

Also I removed an extra `:` from `vim.cmd`.


sindrets:
Good call! I've been meaning to do this and I forgot about it.
Thanks for the PR! :)

PR #70, opened 2021-09-06
feat: Implement enhanced diff highlighting.

### LINKED ISSUE #52: Hooks for tweaking diff highlights
I realise this is a bit of an odd request, but here goes anyway:

Highlighting of diffs in (Neo)Vim is a counter-intuitive. When you have version A and B of a file and diff them, A will show the diff relative to B, while B shows the diff relative to A. In practise this means that e.g. a line removed by B shows up as an addition in A, while not showing up in B. Here's an example of that:

Screenshot from 2021-08-02 18-02-24

In addition, changes in B tend to highlighted using the `DiffText` highlight group. This can result in diffs largely looking similar highlight wise, making it difficult to see what actually changed.

For vim-fugitive I have a hack to fix this (see [here](https://www.reddit.com/r/neovim/comments/ok8u3w/improving_diff_highlighting_when_using_fugitive/)). This basically uses `winhl` to overwrite highlight groups in the appropriate fugitive buffer. The result is something that looks like this:

Screenshot from 2021-08-02 18-05-06

Note how in the old version (left) deletions are shown in red, while additions in the new file (right) are in green. This is what you'd expect when looking at a diff, instead of the default Vim mess.

For diffview.nvim this doesn't work. Buffer names don't have predictable names I can use to inject my `winhl` rules. There are no hooks of any kind that I can tap into either. Which brings me to my request:

Would it be possible to somehow expose a hook that runs for old/new versions in a diff, or perhaps use specific buffer names (e.g. `before: diffview/XXX.rb`) for old/new versions? That would make it possible to fix the diff highlights.

sindrets:
Resolves #52.

As pointed out by @YorickPeterse, the way vim highlights diffs isn't the most intuitive when you're comparing only two files (vim has to do it this way of course, because it's the only way that makes sense when you're comparing *more* than two files).  When comparing old state on the left with new state on the right, what vim highlights as an addition on the left, actually represents a deletion.

This PR adds a new option `enhanced_diff_hl` that attempts to tackle this issue. When enabled, a couple new highlighting groups will be generated based on your colorscheme. These will be applied as local `winhl` in the diff buffers.

```vimhelp
enhanced_diff_hl                            *diffview-config-enhanced_diff_hl*
        Type: `boolean`, Default: `false`

        Enable/disable enhanced diff highlighting. When enabled, |hl-DiffAdd|
        in the left diff buffer will be highlighted as |hl-DiffDelete|, and
        the delete fill-chars will be highlighted more subtly.
```

Before:

2021-09-06-151601_maim

After:

2021-09-06-151538_maim

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ local cb = require'diffview.config'.diffview_callback
 require'diffview'.setup {
   diff_binaries = false,    -- Show diffs for binaries
   use_icons = true,         -- Requires nvim-web-devicons
+  enhanced_diff_hl = false, -- See ':h diffview-config-enhanced_diff_hl'
   signs = {
     fold_closed = "",
     fold_open = "",
--- a/doc/diffview.txt
+++ b/doc/diffview.txt
@@ -7,14 +7,14 @@ Author: Sindre T. Strøm
 
 ==============================================================================
 
-INTRODUCTION                                       *diffview-nvim-introduction*
+INTRODUCTION                                            *diffview-introduction*
 
 Vim's diff mode is pretty good, but there is no convenient way to quickly bring
 up all modified files in a diffsplit. This plugin aims to provide a simple,
 unified, single tabpage interface that lets you easily review all changed files
 for any git rev.
 
-USAGE                                                   *diffview-nvim-usage*
+USAGE                                                        *diffview-usage*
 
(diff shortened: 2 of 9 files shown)


YorickPeterse:
@sindrets Thanks!

PR #71, opened 2021-09-08
Do not change cursor position on select_entry

pappasam:
Both `select_next_entry` and `select_prev_entry` do not change cursor position. To be consistent with this behavior, I think it would be best to not move the cursor when calling `select_entry`.

Alternatively, we can define a new function to preserve existing behavior. That said, I find myself visually browsing diffs more often than I need to move my cursor into the diffs, so I think this is a reasonable default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ for any git rev.
 Install the plugin with your package manager of choice.
 
 ```vim
-" Plug 
+" Plug
 Plug 'sindrets/diffview.nvim'
 ```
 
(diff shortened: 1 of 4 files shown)


sindrets (changes requested):
Thanks for the PR.

I'm fine with making this the default, but I would like to keep the current functionality. Just rename the current event to `focus_entry` or something. You don't need to add a mapping for it. However, please:

- Make the same changes to the equivalent event listeners in the file history view.
- Amend the docs. You can create a new section in the docs under all the mappings for "Available unused maps", where you can document `focus_entry`.

pappasam:
`focus_entry` created and documented!

sindrets (approved):

sindrets:
Many thanks!

PR #79, opened 2021-09-23
fix: file history when .git is not on upper level

### LINKED ISSUE #76: DiffViewFileHistory doesn't work if .git is on upper level
Hello.
DiffViewFileHistory doent work if my working directory is not on same level with .git.
Sometimes I open my project deeper than root level.
Can it be fixed?

farbodsz:
Resolves #76 

Fixes the issue of `DiffviewFileHistory` not working when the project root is not on the same level as the `.git` folder.

The cause of the issue was that paths being passed to the `FileHistoryView` are relative to the vim project root, but not the `.git` project root - e.g. passing `diffview.lua` instead of `lua/diffview.lua`.

### DIFF
--- a/lua/diffview/git/utils.lua
+++ b/lua/diffview/git/utils.lua
@@ -381,6 +381,19 @@ function M.git_dir(path)
   return vim.trim(out)
 end
 
+---Get the path of the given file relative to the project .git directory.
+---@param path string
+---@return string|nil
+function M.git_relpath(path)
+  local out = vim.fn.system(
+    "git ls-tree --full-name --name-only HEAD " .. vim.fn.shellescape(path)
+  )
+  if utils.shell_error() then
+    return nil
+  end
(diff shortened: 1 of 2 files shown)


sindrets:
We can't do it this way because `git ls-tree` does not support pathspec magic (like i.e. `:!exclude/some/path`). Further, this approach does not work with directory paths.

But a very simple way to fix this is to simply use absolute paths when we expand `%` here:
https://github.com/sindrets/diffview.nvim/blob/520bb5c34dd24e70fc063d28bd6d0e8181bff118/lua/diffview/lib.lua#L82

Like:

```lua
table.insert(paths, vim.fn.expand("%:p"))
```

sindrets:
Closing this in favor of 7aa94bcb4eefb07d49a94e1622f845b16f59833f. Regardless: thanks for your work!

PR #84, opened 2021-10-06
feat(file history): add commits hash

### LINKED ISSUE #83: Show commits hashes in DiffviewFileHistory
Is there currently a way to get a commit's hash in `DiffviewFileHistory`?
I think it'd be nice to add the short hash right before the commit's description.

Great plugin though :muscle:

AnatoleLucet:
Closes #83


AnatoleLucet:
> I would like to move the SHA until right before the commit message.

I actually find it really useful there because it's easy to copy since the cursor is already positioned on the hash.
Though I do agree with you that it'd be nice if it was positioned right before the commit message.

Anyway, I added a mapping so the hash can still be copied easily. Let me know what you think :slightly_smiling_face:

PR #107, opened 2022-01-08
Configurable file panel summary

andreypopp:
Add `file_panel.render_summary` boolean option (defaults to `true`) to control if changes summary should be rendered at the top of the file panel.

This makes one less line in the file panel which is really useful for small screens.

Thanks for the plugin!

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ require'diffview'.setup {
     width = 35,                         -- Only applies when position is 'left' or 'right'
     height = 10,                        -- Only applies when position is 'top' or 'bottom'
     listing_style = "tree",             -- One of 'list' or 'tree'
+    render_summary = true,              -- Render changes summary at the top
     tree_options = {                    -- Only applies when listing_style is 'tree'
       flatten_dirs = true,              -- Flatten dirs that only contain one single dir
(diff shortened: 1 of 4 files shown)


sindrets:
@andreypopp Thanks for the PR!

The headings are there to separate sections in the file panel. As this PR currently stands you are only hiding the heading for the working files, and not for the staged files. This creates a weird asymmetry between the sections.

I don't really understand the argument about saving screen space, as I don't think removing one line really makes a notable difference in that regard. This feels more like an aesthetic preference, which is fine. However, whatever change you make to the working files section you also need to make to the staged files section. So I suggest changing the config option to something like `show_headings`, and then you need to find a way to separate the two sections when headings are disabled.

Also, if you want a more compact file list I suggest you set `listing_style` to `list` rather than `tree` if you haven't already. This displays the files as a flat list, which can save a good amount of space.

Let me know if you have any questions!

andreypopp:
Totally missed that it also shows staged diffs as well (which is awesome btw). Ok, I think I can live with the extra lines as it's not 100% obvious to me how to render separate sections without headers.

Thanks for the plugin again, I find it very useful!

PR #126, opened 2022-04-02
fix: Explicitly set foldenable for the diff view

wookayin:
Users may have `set nofoldenable` option in their vimrc, in which case
the diff view windows will not have fold when opening. The `foldenable`
option therefore needs to be explicitly set.


sindrets:
LGTM, thank you!

PR #153, opened 2022-06-11
chore: Add highlight group for WinSeparator.

litearc:
This adds the highlight group for WinSeparator, which has superseded VertSplit. This allows the border between the Diffview panel and the buffer window to be changed without changing the border between buffer windows.


sindrets:
Thanks for the PR! Has WinSeparator fully replaced VertSplit in 0.7.0? If so, please just remove VertSplit entirely.

litearc:
The docs say "hl-VertSplit superseded by hl-WinSeparator", so I think so. I removed VertSplit.

sindrets:
LGTM, thanks!

Issue #154, opened 2022-06-14
`:DiffviewFileHistory` Failing After Update


harrisoncramer:
Hello,

I've recently run `:PackerSync` to get on the latest version of diffview, and now when I'm running the `:DiffviewFileHistory` command (with and without the `%` argument) to get the history of a particular file, it doesn't work. The message is:

```
[Diffview.nvim] No git history for target(s) given the current options! Targets: 'src/components/my-component.vue'
Current options: [ Top-level path: '~/my-repo', Flags: '-n256' '--diff-merges=first-parent' ]
```

This command was working previously without any issue. Has something changed about the API recently that I need to configure this command differently? Didn't see anything mentioned about breaking changes when I updated.

Here's the issue replicated in a totally new repository: https://asciinema.org/a/nKSKF2t4NY7OGQasandW12sud

sindrets:
@harrisoncramer there haven't been any breaking changes since #151. What version of Git are you running?

harrisoncramer:
I'm on `2.30.2`

harrisoncramer:
FWIW this is likely the change that is causing me issues because I infrequently update my Packer dependencies. Am I doing something wrong in that ASCIICinema?

sindrets:
@harrisoncramer It seems my intuition was correct: your git version is too old. The `--diff-merges` option for git-log was added in git 2.31.0. Try to update your git to a more recent version.

wongjiahau:
@sindrets, is it possible to use the old git command if the git version is not up-to-date?

Because unfortunately for me, my workstation uses Ubuntu, and unlike Arch, it's not always possible for me to install the latest packages out there. 

Edit: I found a way to download the latest `git` on Ubuntu (https://askubuntu.com/a/568596/788130). 

Regardless, I think it is a breaking change to depend on new APIs of existing dependencies.

Anyway, thanks for your hard work!

PR #157, opened 2022-06-16
Specify git version in README

harrisoncramer:
Relates to #154. The `:DiffviewFileHistory` command will not work prior to this version of Git.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ for any git rev.
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


gegoune:
Perhaps adding healthcheck could help as well?

sindrets:
@gegoune yes, either that or we can check support once before running any related git jobs. I'll look into it when I get some time.

Issue #159, opened 2022-06-18
Vim(filetype):E21: Cannot make changes, 'modifiable' is off: fileencoding=utf-8


cathaysia:
a bug just like this:

https://user-images.githubusercontent.com/42114817/174433325-e8e7885e-0466-4ab1-ba77-723f5a9c33f3.mp4

And this can be fix by set `vim.bo[bn].modifiable = true`

diffview.nvim/lua/diffview/views/file_entry.lua:412

PR #160, opened 2022-06-18
fix: E21: Cannot make changes, 'modifiable' is off

cathaysia:
This example codes to fix https://github.com/sindrets/diffview.nvim/issues/159. So merge it or not up to you. :)

### DIFF
--- a/lua/diffview/views/file_entry.lua
+++ b/lua/diffview/views/file_entry.lua
@@ -253,7 +253,7 @@ function FileEntry:load_buffers(git_root, left_winid, right_winid, callback)
 
   self.left_bufid = splits[1].bufid
   self.right_bufid = splits[2].bufid
-  vim.cmd("do WinEnter")
+  pcall(vim.cmd, "do WinEnter")
 
   perf:lap("load done")
 end
@@ -412,8 +412,8 @@ function FileEntry._create_buffer(git_root, rev, path, null, callback)
           vim.api.nvim_buf_call(bn, function()
             vim.cmd("filetype detect")
           end)
-          callback()
           vim.bo[bn].modifiable = false
+          callback()
         end
       end)
     else
@@ -493,7 +493,7 @@ function FileEntry._update_windows(left_winid, right_winid)
           -- consistently than calling `:syncbind`.
           vim.cmd([[exe "norm! \<c-e>\<c-y>"]])
         end
-        vim.cmd("do WinLeave")
+        vim.cmd("do <nomodeline> WinLeave")
       end)
     end
   end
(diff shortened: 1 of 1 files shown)


sindrets (changes requested):
Thanks for PR! The buffers created from git history should remain un-modifiable. The root of the problem here is that the `:h modeline` in the file is trying to set the fileencoding of an un-modifiable buffer. I think we should just `pcall` the call to `filetype detect` instead, to catch the error.

cathaysia:
In fact, if I remove https://github.com/sindrets/diffview.nvim/blob/1eb1e497e36eea455d85b7e83e4272222a882ffa/lua/diffview/views/file_entry.lua#L414 , another error will happen.

emm, this bug disappeared on my laptop. I cannot test this bug. So should we close this pr and https://github.com/sindrets/diffview.nvim/issues/159?

sindrets:
@cathaysia Do not remove `filetype detect`. Just pcall it.

> another error will happen.

Yes, probably `do WinEnter` on L256 which will re-trigger the modeline. Just pcall that as well.

> emm, this bug disappeared on my laptop. I cannot test this bug.

Like I said, this is caused by modelines. You can easily create a repo that reproduces this with a simple script like:

```bash
#!/usr/bin/bash

rm -rf /tmp/pr160-repo
mkdir /tmp/pr160-repo
cd /tmp/pr160-repo
git init
echo ' vim:fileencoding=utf-8' > a
git add -A
git commit -am 'init'
```

Then open the `DiffviewFileHistory` for this repo, and you will see the error.

cathaysia:
emm, I had give a try on pacll, this bug can be solved by `vim.schedule(pcall....`, but it will introduce another bug: if pcall failed, diffview will be blank when call DiffviewFileHistory:

图片

the codes are:

```lua
      vim.schedule(function()
        pcall(function()
          if api.nvim_buf_is_valid(bn) then
            vim.bo[bn].modifiable = true
            api.nvim_buf_set_lines(bn, 0, -1, false, result)
            vim.bo[bn].modifiable = false
            vim.api.nvim_buf_call(bn, function()
              vim.cmd("filetype detect")
            end)
            callback()
          end
        end)
      end)
```

So, may move move line 414 down be a better way:

```lua
          vim.bo[bn].modifiable = true
          api.nvim_buf_set_lines(bn, 0, -1, false, result)
          vim.api.nvim_buf_call(bn, function()
            vim.cmd("filetype detect")
          end)
          callback()
          vim.bo[bn].modifiable = false
```

cathaysia:
In my opinion, set modifiable to false just to prevent user to modify diffview, so we can put this last.

cathaysia:
Yes, what you say is the best solution.

I had give a try to move vim.bo[bn].modifiable = false down, or pcall(callback), but then diffview cannot deal with error modeline, the best solution is close modeline. :)

test code:

``` bash
#!/usr/bin/bash

rm -rf /tmp/pr160-repo
mkdir /tmp/pr160-repo
cd /tmp/pr160-repo
git init
echo ' vim:fileencoding=utf-16' > a
git add -A
git commit -am 'init'
```

Issue #22, opened 2021-05-16
Merge Conflicts


TimUntersberger:
This [stackoverflow question](https://stackoverflow.com/questions/14904644/how-do-i-use-vimdiff-to-resolve-a-git-merge-conflict) is a great starting point for understanding how merge conflict resolution works with the builtin diffing tool.

One thing I noticed right away is that by default 4 splits appear. I'm not really a fan of this kind of resolution so I would propose either a traditional 2 or 3 way split approach. We could also try to replicate VSCode's way of doing this as it is very space efficient and based on my experience it is not too confusing.

We would also need to think about how we could reshape the API to support both normal diffing and merge conflict resolution.

@sindrets have you thought about how you want to implement this before?

ur4ltz:
Hi!

See this plugin [vim-mergetool](https://github.com/samoshkin/vim-mergetool) and video [Vim as a merge tool](https://www.youtube.com/watch?v=VxpCgQyUXlI).

Best regard Andy.

sindrets:
@TimUntersberger I don't have anything planned out yet. But like you say, the first thing we need to figure out is the layout. How many splits do we need? I agree with you that we should reduce the number of splits needed if we can. How do we want the workflow of selecting one of the conflict hunks to work? (I think I want to be able to use `:diffget` to do this).

In terms of implementation, I think we should create a new subclass of `View` that handles the mergeview layout. The file panel should work just fine for listing the conflicting files. And we probably need to refactor `FileEntry` to be able to accommodate both types of views.

@ur4ltz Thanks, I've actually already seen it, but it's a good reference for how it can be done!

rafamadriz:
Just as a reference, here's another really good video on how to resolve merge conflicts with (neo)vim, http://vimcasts.org/episodes/fugitive-vim-resolving-merge-conflicts-with-vimdiff/

As @sindrets mentioned, `diffget` is used along with `Gdiff` from fugitive, the thing is that fugitive uses a specific naming convention for the buffers when you use Gdiff, this makes it easier to identify buffers when doing `diffget`

serhez:
Hello! Any updates on this enhancement? It would be great to not depend on `fugitive` just for this one feature, which is the only plugin I've found that provides a 3-way split etc. Thanks!

sindrets:
[...]

PR #190, opened 2022-08-18
refactor!: Flexible layout system.

sindrets:
This is a huge refactor that gives the layout system in the views far greater flexibility. The primary motivation behind this PR is the planned merge-tool (#22) that is going to make use of 3-way diffs. I want the conflicted files to appear in their own section in the normal Diffview, thus the views need to be able to restructure the layout dynamically depending on what entry is currently open. The current implementation is deeply tied to the 2-way diff layout, and it would be unfeasible to force it to conform to a bodged dynamic layout system without larger structural changes.

#### The only breaking change affects the action `view_windo`. This action is not in use by default, so you won't be affected unless you know that you've made use of it in your config.

Different entries in the same view may now use completely different window layouts. Thus the action `view_windo` has changed to reflect these changes. See `:h diffview-actions-view_windo` for more details on the new usage.


melkster:
@sindrets After updating to the latest version of Diffview I now get the error `packer.nvim: Error running config for diffview.nvim: ...site/pack/packer/start/diffview.nvim/lua/diffview/hl.lua:240: invalid key: underlineline` whenever I launch Neovim 🤔

sindrets:
@melkster What Neovim version are you on? The highlight attribute `underlineline` was added in March quite a while before the v0.7.0 release (which is the minimum required version for this plugin, as stated in the README).

melkster:
I'm on nightly `NVIM v0.8.0-dev+580-g45ba2e147`, but I haven't updated in a while so maybe there's an issue with the commit I'm on

sindrets:
Ah, I see that the attribute has been renamed in nightly: https://github.com/neovim/neovim/commit/995e4879153d0f4ea72dff446c175754a1873425

sindrets:
@melkster Should be fixed now in the latest commit.

melkster:
@sindrets Thanks!

PR #203, opened 2022-09-02
allow setting local opts via new 'win_opts' config property

### LINKED ISSUE #178: [feature] Allow specifying window-local options in `win_config`
To allow customisation of the stausline in the file panel win

e.g. 

```lua
{
  type = "split",
  statusline = " "
}
```

..would essentially disable the statusline for the file panel 

--

I can have a crack at implementing this if I get a sec

ceuk:
Resolves #178.

Allows users to set local options via a `win_opts` property in each respective `win_config` table. e.g.

```lua
require("diffview").setup({
   file_panel = {
      win_config = {
         win_opts = {
            number = true,
            statusline = "hello, statusline"
         }
      }
   }
})
```


lua/diffview/ui/panel.lua
```diff
@@ -288,7 +275,7 @@ function Panel:open()
   end
 
   self:resize()
-  utils.set_local(self.winid, self:class().winopts)
+  utils.set_local(self.winid, config.winopts)
```
sindrets, line 278: Thanks for the PR!

I think we can simplify this a lot by leaving the window options defined in the various panel classes, and treat `win_config.winopts` as overrides. Just apply the `win_config` window options after the classes own options:


```suggestion
  utils.set_local(self.winid, self:class().winopts)
  utils.set_local(self.winid, config.winopts)
```

---

Could you also please [amend the config validation](https://github.com/sindrets/diffview.nvim/blob/e32de2be00e0b5e23736f906f3d6d9dca14fbeb0/lua/diffview/ui/panel.lua#L160-L202) to ensure that `winopts` is a table (`:h vim.validate()`)?

I know I was the one who suggested the name `winopts`, but let's rename it to `win_opts`. Otherwise it would be strangely inconsistent with the naming of all the other config options.

sindrets (changes requested):

lua/diffview/ui/panel.lua
ceuk: Great shout! Can't believe how much simpler that's made it :)

doc/diffview.txt
```diff
@@ -73,10 +74,13 @@ Example configuration with default settings:
         win_config = {    -- See |diffview-config-win_config|
           position = "bottom",
           height = 16,
+          win_opts = {}
         },
       },
       commit_log_panel = {
-        win_config = {},  -- See |diffview-config-win_config|
+          win_config = {    -- See ':h diffview-config-win_config'
```
sindrets, line 81: Please use `|diffview-config-win_config|` here so that users can jump directly to the tag with `:h CTRL-]`.

doc/diffview.txt
```diff
@@ -213,6 +217,11 @@ win_config                                      *diffview-config-win_config*
                 current window.
                 Default: `0`
 
+            {win_opts} (table)
+                Any vim options to set for the local window (see |vim.opt_local|).
+                You don't have to copy all the defaults, just specify any
+                values you want to set/override.
```
sindrets, line 223: > You don't have to copy all the defaults ...

This doesn't really make sense now that the defaults are empty. Maybe change this paragraph to something like:

```vimhelp
Table of window local options (see |vim.opt_local|). These options are applied
whenever the window is opened.
```

sindrets (changes requested):
Very nice! Just a couple of small things, then I'll merge :)

ceuk:
Hey, apologies for the delay, I've made the other changes and re-aligned the docs with the new stuff on main. Should be good to go I think :+1:

sindrets (approved):
Thanks!

PR #209, opened 2022-09-08
Fix #206 showing of untracked files

### LINKED ISSUE #206: [Bug] Diffview ignores status.showUntrackedFiles
Hi, I really love your plugin, thank you for your time and effort.

For my dotfiles I use a bare repository in $HOME. When editing dotfiles I have a script which sets `GIT_DIR=$HOME` and `GIT_WORK_TREE=$HOME/.dotfiles` which gets picked up by nvim / diffview.nvim correctly.

But when opening diffview.nvim it takes ages to load (and blocks nvim while loading) due to the plugin trying to show all untracked files in my $HOME which are over 500k files.
In git I have set  `git config status.showUntrackedFiles no` to prevent a git diff/status showing all files.
I'm aware I could call `:DiffViewopen -uno` to prevent showing all untracked files (which works), but it would be great if diffview.nvim would pickup `status.showUntrackedFiles` from the git environment automatically.

Actually I debugged your code and found out that the function `show_untracked(..)` in `lua/diffview/git/utils.lua` returns true when opening the diffview (as the call to `exec_sync(..)` returns with an 128 error and `out` being nil). If I just `return false` from the function it works as I would expect.

If I change the function like so, everything works for me:
```
function M.show_untracked(toplevel)
    local out = M.exec_sync(
        { "config", "status.showUntrackedFiles" },
        { cwd = toplevel, silent = true }
    )
    return vim.trim(out[1] or "") ~= "no"
end
```
Maybe this helps?

splitDEV:
Fixes #206

Treat `git config status.showUntrackedFiles` not as bool.

### DIFF
--- a/lua/diffview/git/utils.lua
+++ b/lua/diffview/git/utils.lua
@@ -1554,10 +1554,10 @@ end
 ---@return boolean
 function M.show_untracked(toplevel)
   local out = M.exec_sync(
-    { "config", "--type=bool", "status.showUntrackedFiles" },
+    { "config", "status.showUntrackedFiles" },
     { cwd = toplevel, silent = true }
   )
-  return vim.trim(out[1] or "") ~= "false"
+  return vim.trim(out[1] or "") ~= "no"
 end
 
 ---Get the diff status letter for a file for a given rev.
--- a/lua/diffview/lib.lua
+++ b/lua/diffview/lib.lua
@@ -86,8 +86,8 @@ function M.diffview_open(args)
     show_untracked = arg_parser.ambiguous_bool(
       argo:get_flag({ "u", "untracked-files" }, { plain = true }),
       nil,
-      { "all", "normal", "true" },
-      { "no", "false" }
+      { "all", "normal" },
+      { "no" }
     ),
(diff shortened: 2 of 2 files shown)


splitDEV:
I was unsure if the changes to [lua/diffview/lib.lua](https://github.com/sindrets/diffview.nvim/pull/209/files#diff-cbef83aeff6fca3acd62507ff6af85c1efe23b296ed500dbc23b180c14499778) are necessary? It is the correct handling of status.showUntrackedFiles but maybe you did have some use cases where you needed to treat it as bool with false/true?

sindrets:
Yes, please don't change the argument handling in `lib.lua`. The `-u` in diffview is actually a boolean flag. You can see I'm referring to it as an "ambiguous bool". I've just mapped additional values to true and false such that it will be familiar to people who are used to the flag option from git. The arg parser will automatically assign the value `true` to flags that aren't given a value by the user. Essentially `--staged` becomes `--staged=true`. You can also see that the completion will suggest the values `true` and `false` if you request completion for any of the boolean flags.

splitDEV:
Ok, reverted the change

sindrets (approved):
LGTM, thanks!

PR #216, opened 2022-09-17
added option to disable auto refresh

### LINKED ISSUE #208: [Feature] Configurable automatic refreshing
When I stage a hunk on `DiffviewOpen` the diff is refreshing and its annoying, you have any idea why it happens and how can I stop it?

ofirgall:
Resolves #208.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ require("diffview").setup({
   enhanced_diff_hl = false, -- See ':h diffview-config-enhanced_diff_hl'
   git_cmd = { "git" },      -- The git executable followed by default args.
   use_icons = true,         -- Requires nvim-web-devicons
+  watch_index = true,       -- Update views and index buffers when the git index changes.
   icons = {                 -- Only applies when use_icons is true.
     folder_closed = "",
     folder_open = "",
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -154,6 +154,7 @@ require("diffview").setup({
   enhanced_diff_hl = false, -- See ':h diffview-config-enhanced_diff_hl'
   git_cmd = { "git" },      -- The git executable followed by default args.
   use_icons = true,         -- Requires nvim-web-devicons
+  auto_refresh = true,      -- Auto refresh the status of the current file
```
sindrets, line 157: Let's call it `watch_index`. And I think we should make the description a little bit more accurate:

```suggestion
  watch_index = true,       -- Update views and index buffers when the git index changes.
```

lua/diffview/scene/views/diff/diff_view.lua
```diff
@@ -124,15 +125,17 @@ function DiffView:post_open()
 
   self.watcher = vim.loop.new_fs_poll()
   ---@diagnostic disable-next-line: unused-local
```
sindrets, line 127: Please move both of these lines inside the if clause you created. No need to create and hold on to a file descriptor we aren't using.

sindrets (changes requested):
Thanks for the PR! I left some comments.

ofirgall:
Made the changes

sindrets (approved):
LGTM, thanks!

PR #229, opened 2022-10-03
fix: Remove unneeded space when not using icons

### LINKED ISSUE #227: use_icons = false still displays spaces used to separate icons in file panel
When using the option use_icons = false, the file panel will display an extra space in between the file status and the file name.

space

After editing the diff/render.lua:

Video showing the difference:

https://user-images.githubusercontent.com/14960414/193579913-13705b12-9508-4a18-8b7d-c0d697b17950.mp4

This probably happens in other places where icons are used too, I haven't checked everything.

charbelnicolas:
Resolves #227.

DiffviewOpen before:

DiffviewOpen after:

DiffviewFileHistory before:

DiffviewFileHistory after:

### DIFF
--- a/lua/diffview/hl.lua
+++ b/lua/diffview/hl.lua
@@ -311,7 +311,7 @@ function M.hi_clear(groups)
 end
 
 function M.get_file_icon(name, ext, render_data, line_idx, offset)
-  if not config.get_config().use_icons then return " " end
+  if not config.get_config().use_icons then return "" end
 
   if not web_devicons then
     local ok
--- a/lua/diffview/hl.lua
+++ b/lua/diffview/hl.lua
@@ -324,7 +324,7 @@ function M.get_file_icon(name, ext, render_data, line_idx, offset)
           .. "Set `use_icons = false` in your config to stop seeing this message."
       )
 
-      return " "
+      return ""
     end
   end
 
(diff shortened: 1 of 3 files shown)


sindrets (changes requested):
The underlying issue seems to be that we're returning whitespace from the `get_file_icon()` function here: https://github.com/sindrets/diffview.nvim/blob/6baa30d0a6f63da254c2d2c0638a426166973976/lua/diffview/hl.lua#L313

If we just return an empty string instead, when `use_icons` is disabled, then that should account for all current and future uses of icons in the interface.

charbelnicolas:
Sorry for polluting the commit history, I'm not very good at git, it probably might just be easier if you do it yourself seeing how easy was the fix

sindrets (changes requested):
This one should also return an empty string:

https://github.com/charbelnicolas/diffview.nvim/blob/41672f3e720e78d7d8dfde9b794ff335ef928102/lua/diffview/hl.lua#L327

> Sorry for polluting the commit history, I'm not very good at git, it probably might just be easier if you do it yourself seeing how easy was the fix

Don't worry, I'll just squash the commits before merging :)

charbelnicolas:
> This one should also return an empty string:
> 
> https://github.com/charbelnicolas/diffview.nvim/blob/41672f3e720e78d7d8dfde9b794ff335ef928102/lua/diffview/hl.lua#L327
> 
> > Sorry for polluting the commit history, I'm not very good at git, it probably might just be easier if you do it yourself seeing how easy was the fix
> 
> Don't worry, I'll just squash the commits before merging :)

Oops, totally missed that one -_-

Issue #233, opened 2022-10-12
Support for other VCS


zegervdv:
Hi, 
I really like the functionality and UI of this plugin, especially for large repositories it is very useful to quickly get the history of the current file in a nice and clear overview. 

Sadly most of my work is in mercurial repos and I would love to get a similar experience there. 

I spent a couple of hours trying to get familiar with the code base and try to hack in mercurial support, but I think as it stands, the code is very much hardwired for git and it would require a large refactor to enable support for other VCS’es. 

So my question is: would you be open to take on that burden of supporting multiple VCS’es (mercurial in particular)?
If so, I am willing to do (some of) that work with your guidance. I guess you probably have some ideas on how you would prefer to implement such a change. 
At the very least I can help out with the equivalent mercurial commands to get the required data, but I’m definitely willing to help out more. 

Either way, thanks for your time.

sindrets:
@zegervdv Thank you for taking an interest in the project!

My stance on including support for other VCS' is that: I don't personally use any other VCS than git, and I'm not interested nor capable of working on such functionality myself (without spending a lot of time familiarizing myself with VCS software that I don't use). However, I'm willing to include it if it's developed and maintained by other people.

It's also fine if it doesn't have full feature parity with what we're offering for git. We can just add a disclaimer that mercurial is only partially supported (if that's the case). I don't know how well i.e. the git-log options translate to mercurial. But I'm guessing that it offers at least similar things.

Is this something you would be interested in working on?

zegervdv:
I definitely understand you don’t want to take on extra work for something you wouldn’t use yourself, which is why I wanted to ask before diving deeper into it. 
I would like to work on it and keep supporting it. 

As for the feature parity, I think it should be possible to get the core functionality matched, probably with some different options or flags. 
As a first goal I’d like to get DiffviewFileHistory working. 

Practically I think this will probably take a couple of PRs. Currently I think the git module is a combination of an abstraction layer and git specific code and that should probably be broken up into a common base and VCS-type specific module using the same interface. Based on which kind of repo you are in the right specific module (git or hg) could then be loaded. 

Does that sound reasonable? Do you have any suggestions or ideas how you prefer to tackle this?

sindrets:
> I would like to work on it and keep supporting it.

@zegervdv Brilliant, thanks!

> Currently I think the git module is a combination of an abstraction layer and git specific code and that should probably be broken up into a common base and VCS-type specific module using the same interface.

[...]

PR #235, opened 2022-10-15
refactor: Refactor git module into vcs-type agnostic module

zegervdv:
Part 1 of adding support for other VCS'es (#233).

This creates a `vcs` module, which replaces the `git.utils` module. Upon requiring it, the vcs type is determined and the right module is returned (currently hardcoded to always return `git`).

I've changed the `Commit` class into a base class and made a `GitCommit` class for git-specific stuff.
The `rev` module will also need to be split, but this will probably will have to be done similarly to how the `vcs` module automagically returns the correct implementation.

Still to do:
- [ ]  handle the rev module
- [ ] move common parts from `git.utils` to `vcs.utls` (all the job handling etc)

@sindrets is this in line with your expectations?


sindrets (commented):
EDIT: I started typing out this comment before you pushed the latest changes, but I think most of it still applies.

---

I think we should probably create a VCS adapter interface that all the VCS adapters implement. The VCS module should not return an adapter on require. The appropriate adapter needs to be determined from a set of paths. Which is why i referred to the `find_git_toplevel()` function previously. I think the way we implement adapter acquisition should be similar to how this function works: we give it an ordered list of paths that might indicate what repository we are part of, and then it goes through the list and checks if any of these paths are part of a repository of any of the supported VCS'.

Making the adapters more object oriented than what the current git module is, is a good call I think. Currently pretty much all the git functions requires a reference to either one, or both of the git top-level and the git dir. These are context variables that I think we can safely assume will not change for the lifetime of a view, and thus we would benefit from storing these in the adapter instances (I don't know if Mercurial uses the equivalent of a git dir, but it necessarily needs the top-level path regardless).

This change is going to require a fair amount of refactoring. We need to instantiate the adapters when we create views in `lib.lua`, and then pass these adapters to the view constructors. All the subcomponents of a view will acquire the adapter through their parent view.

Refactoring the project to use this model should be the focus of this first PR I think. From there it seems pretty straight forward to implement other VCS adapters.

lua/diffview/vcs/init.lua
```diff
@@ -1,3 +1,7 @@
-local git = require('diffview.vcs.adapters.git.utils')
+local M = {}
 
-return git
+function M.get_adapter(path)
```
zegervdv, line 3: Idea is that the lib calls this `get_adapter` with the current path, this is used to determine which kind of repo we are in.

For now I will make this return the `GitAdapter` so the existing features can be restored first.

zegervdv:
So, instead of `local git = require("diffview.git.utils")` we do something like:
``` lua
local vcs = require("diffview.vcs")
local adapter = vcs.get_adapter(path)
```
once, and pass that created `adapter` object around to all functions?

For example, in `init.lua` there are a few calls to `git.git_dir`. These would then need to be called on the adapter instantiation, I'm not sure it you want all those functions to pass around this object?

sindrets:
The views are already passing around a `GitContext` object, because it holds data that is necessary to run any git command. We could just pass an adapter instance instead that would hold this same data, as well as offering the VCS interface. This is actually simpler than what I'm currently doing.

```lua
-- Instantiate adapter:
local adapter = require("diffview.vcs").get_adapter(paths)

-- Now the adapter instance would have access to all the necessary context data
-- for the repo derived from the given paths:
adapter.ctx.toplevel
adapter.ctx.git_dir
```

---

> For example, in `init.lua` there are a few calls to `git.git_dir`. These would then need to be called on the adapter instantiation, I'm not sure it you want all those functions to pass around this object? 

The git stuff in `init.lua` is all related to offering git completion. This would have to be moved into the adapters regardless. So in these cases the adapter instance approach is simpler as well. Because currently, in all these cases, we're currently deriving both the git top-level and the git dir because this data is a prerequisite to run the necessary git commands.

So instead of doing this:

```lua
local toplevel = lib.find_git_toplevel(paths)
if toplevel then
  local git_dir = git.git_dir(toplevel)
  if git_dir then
    git.thing_we_actually_need(toplevel, git_dir)
  end
end
```

We would do this:

```lua
local adapter = vcs.get_adapter(paths)
if adapter then
  adapter.thing_we_actually_need()
end
```

zegervdv:
I see, I was missing the scale of the refactoring here :)

So to summarize:
 - create a VCSAdapter class which will define the agnostic interface and implement some common functionality (exec_sync etc) now in `git.utils` and extend with init and lib functions
 - implement the GitAdapter with code from `git.utils` and  init/lib
 - instantiate the adapter

I'll be pushing some commits whenever I find some time, when I have something worth reviewing I'll ping you again.

zegervdv:
@sindrets 

### A status overview of what I've been doing:

- Created a `VCSAdapter` class in `vcs/adapter.lua`, this defines the generic interface and implements some common code
- `vcs/init.lua` is where the `get_adapter` factory is defined
- `vcs/adapers/git/init.lua` holds the `GitAdapter` implementation and the flags for the `FHOptionPanel`
- `vcs/utils.lua` has most of the job handling code

Currently, the `DiffviewFileHistory` command is working again with default config, and at least the arguments I tried with seem to still work fine.

### Still TODO
- Fix the other commands/features
- Rename the git-specific fields and names to something more generic
- Clean up or add the annotations
- Remove leftover or duplicated code
- Some further refactoring of the `GitAdapter` to move out some common parts will probably become clearer when implementing the `HgAdapter` variant
- Sort out the completion handling

### Questions

#### Instantiation
I'm a bit struggling to find the right place to instantiate the adapter because I seem to get into a chicken-or-egg problem:
I want to figure out which repo type we are in, but also support arguments (with completion) and paths.  But those paths may change which directory we are looking at and therefore change the repo-type.

For example, I'm working in a git repo which has a submodule repo. If I open nvim from the top repo directory and navigate to one of the submodule files and run `DiffviewFileHistory` there, I think the desired behavior would be to open the history of the submodule, or do I see that wrong? 

But if that is the case, the adapter can only be created when a path or a `-C` option is given, which means I can't provide repo specific completions.
I'm not really sure what the best way forward is:
- separate commands per repo type?
- do a best effort guess based on buffer and cwd, which means doing a `DiffviewFileHistory path/to/submodule/file` will not produce the correct result
- other options?

Currently I am parsing the args as git first, giving back the paths and extracted toplevel indicators, which are then used to create the adapter, which sets the `toplevel` and `(git)dir` based on those indicators.


#### Config
We'll probably need to split some config options into common and `git` specific fields. How do you wish to handle that?
If the current structure changes, we will need to provide proper deprecation warnings so users know how to adapt their config.

sindrets:
@zegervdv Thank you for doing this thoroughly!

I see the chicken-or-egg problem you're facing. It's a bit tricky, but I think I've found a good way to solve it. Please take a look at this suggested implementation:

<details>
  <summary><code>vcs/adapters/git/init.lua</code></summary>

```lua
---@param path_args string[] # Raw path args
---@param cpath string? # Cwd path given by the `-C` flag option
---@return string[] path_args # Resolved path args
---@return string[] top_indicators # Top-level indicators
function M.get_repo_paths(path_args, cpath)
  local paths = {}
  local top_indicators = {}

  for _, path_arg in ipairs(path_args) do
    for _, path in ipairs(pl:vim_expand(path_arg, false, true)) do
      local magic, pattern = pathspec_split(path)
      pattern = pl:readlink(pattern) or pattern
      table.insert(paths, magic .. pattern)
    end
  end

  local cfile = pl:vim_expand("%")
  cfile = pl:readlink(cfile) or cfile

  for _, path in ipairs(paths) do
    if pathspec_split(path) == "" then
      table.insert(top_indicators, pl:absolute(path, cpath))
      break
    end
  end

  table.insert(top_indicators, cpath and pl:realpath(cpath) or (
    vim.bo.buftype == ""
    and pl:absolute(cfile)
    or nil
  ))

  if not cpath then
    table.insert(top_indicators, pl:realpath("."))
  end

  return paths, top_indicators
end
```
</details>

<details>
  <summary><code>vcs/init.lua</code></summary>

```lua
---@class vcs.init.get_adapter.Opt
---@field top_indicators string[]?
---@field cmd_ctx vcs.init.get_adapter.Opt.Cmd_Ctx? # Context data from a command call.

---@class vcs.init.get_adapter.Opt.Cmd_Ctx
---@field path_args string[] # Raw path args
---@field cpath string? # Cwd path given by the `-C` flag option

---@param opt vcs.init.get_adapter.Opt
---@return err string?
---@return adapter VCSAdapter?
function M.get_adapter(opt)
  local adapters = { git, hg }

  for _, adapter in ipairs(adapters) do
    local path_args
    local top_indicators = opt.top_indicators

    if not top_indicators then
      path_args, top_indicators = adapter.get_repo_paths(opt.cmd_ctx.path_args, opt.cmd_ctx.cpath)
    end

    local toplevel = adapter.find_toplevel(top_indicators)

    if toplevel then
      -- Create a new adapter instance. Store the resolved path args and the
      -- cpath in the adapter context.
      return nil, adapter.create(toplevel, path_args, opt.cmd_ctx.cpath)
    end
  end

  return "Not a repo (or any parent), or no supported VCS adapter!"
end
```
</details>

<details>
  <summary><code>vcs/adapters/git/init.lua</code></summary>
  
```lua
---@param toplevel string
---@param path_args string?
---@param cpath string?
function M.create(toplevel, path_args, cpath)
  return GitAdapter({
    toplevel = toplevel,
    path_args = path_args,
    cpath = cpath,
  })
end

function GitAdapter:init(opt)
  opt = opt or {}
  GitAdapter:super().init(self, opt)

  self.bootstrap.target_version = {
    major = 2,
    minor = 31,
    patch = 0,
  }

  local cwd = opt.cpath or vim.loop.cwd()

  self.ctx = {
    toplevel = opt.toplevel,
    git_dir = self:git_dir(opt.toplevel),
    path_args = vim.tbl_map(function(pathspec)
      return pathspec_expand(opt.toplevel, cwd, pathspec)
    end, opt.path_args or {}) --[[@as string[] ]]
  }
end
```
</details>

<details>
  <summary><code>lib.lua</code></summary>

```lua
function M.file_history(range, args)
  local default_args = config.get_config().default_args.DiffviewFileHistory
  local argo = arg_parser.parse(vim.tbl_flatten({ default_args, args }))
  local rel_paths

  local err, adapter = vcs.get_adapter({
    cmd_ctx = {
      path_args = argo.args,
      cpath = argo:get_flag("C", { no_empty = true, expand = true }),
    },
  })

  if err then
    utils.err(err)
    return
  end

  rel_paths = vim.tbl_map(function(v)
    return v == "." and "." or pl:relative(v, ".")
  end, adapter.ctx.path_args)

  -- The rest remains pretty much the same, except it doesn't need to find the
  -- git dir, and we can now use the adapter instance to perform the dry-run,
  -- option verification etc...
end
```
</details>

<details>
  <summary><code>init.lua</code></summary>

```lua
-- For the completion we can now still get an adapter by just explicitly
-- setting the top_indicators.

M.completers = {
  ---@param ctx CmdLineContext
  DiffviewOpen = function(ctx)
    local cfile = utils.path:vim_expand("%")
    local top_indicators = utils.vec_join(
      vim.bo.buftype == ""
          and utils.path:absolute(cfile)
          or nil,
      utils.path:realpath(".")
    )

    -- Create a short-lived adapter such that we can get relevant completion
    local err, adapter = vcs.get_adapter({ top_indicators = top_indicators })

    -- ...
  end
}
```
</details>

---

> For example, I'm working in a git repo which has a submodule repo. If I open nvim from the top repo directory and navigate to one of the submodule files and run `DiffviewFileHistory` there, I think the desired behavior would be to open the history of the submodule, or do I see that wrong?

That is how it should currently work. I think you just messed up the handling of the toplevel indicators a little bit:

https://github.com/zegervdv/diffview.nvim/blob/dfb733fc901668c39a8b8f4a6da58fec901a2a9f/lua/diffview/vcs/adapters/git/init.lua#L84-L98

Here you're determining that we aren't in a git repo after only checking the `-C` path or the cwd. This is wrong, as we need to check all the top indicators in order. And this is `find_toplevel()`'s job. So this code should not be here at all (except for the part that adds the cwd to the top indicators when `-C` is not present).

> But if that is the case, the adapter can only be created when a path or a -C option is given, which means I can't provide repo specific completions.
> I'm not really sure what the best way forward is:

My suggested implementation should solve all these problems, but let me know if anything is unclear!

> We'll probably need to split some config options into common and git specific fields. How do you wish to handle that?

Yes, we'll have to deal with this, but not right now. This PR should just focus on implementing the VCS adapter model, and then we can address this in the next PR where you actually implement the mercurial adapter.

> If the current structure changes, we will need to provide proper deprecation warnings so users know how to adapt their config.

You can take a look at how I'm handling deprecations currently:

https://github.com/sindrets/diffview.nvim/blob/a1fbcaa7e1e154cfa793ab44da4a6eb0ae15458d/lua/diffview/config.lua#L327

---

You're doing great work here! Just let me know when you would like a proper code review, and I'll find the time :)

zegervdv:
With those suggestions I can indeed get it to work, thanks!
Things are beginning to fall into place.

I think it makes sense that I first work through the TODO's I listed last time before doing a proper review, so you don't have spent too much time on review cycles. 
I'll take the PR out of draft when I think I'm ready for that.

zegervdv:
@sindrets 
I think I'm almost there: `DiffviewFileHistory` and `DiffviewOpen` seem to work again, as far as the arguments that I tested. But there are so many options, I'm sure I haven't covered all possibilities :).

The merge view also seems to work, but I'm not sure about the cleanest implementation there.
I moved the `parse_conflicts` to the common `vcs.utils` because I think most VCS'es use the same format (at least the one without the BASE part, which I believe your implementation also supports?). If needed that could also be moved to the VCSAdapter, but then the `actions` would need to either take the adapter as an argument, or move to the VCSAdapter too (the latter doesn't seem right).

I found 2 functions that don't seem to be called anywhere:
* `get_file_status`
* `get_file_stats`

I've removed these, but maybe you were planning to use these for some new features?

Currently I have a `vcs/init.lua` and a `vcs/utils.lua` which are both imported as `vcs` in different files, I would probably be cleaner to have everything in the `init.lua` file?

I want to do a final pass through the code to make sure I've updated all old function calls and clean up the annotations.
So in the next couple of these it should be ready for a formal code review.

sindrets:
@zegervdv Brilliant! I'll see if I can find the time to review this soon.

> (at least the one without the BASE part, which I believe your implementation also supports?)

Yes, it does :)

`parse_conflicts` can remain in utils until there's a reason to move it.

> I found 2 functions that don't seem to be called anywhere:

I believe they are just left over from previous refactors, so removing them is probably the right thing to do. But I'll take a closer look.

lua/diffview/vcs/adapters/git/init.lua
```diff
@@ -0,0 +1,1717 @@
...
+    "[^.]%.%.%.?$",
+    "[^.]%.%.%.?[^.]",
+    "^.-%^@",
+    "^.-%^!",
+    "^.-%^%-%d?",
+  }) ~= nil
+end
+
+---Parse a given rev arg.
+---@param git_toplevel string
```
sindrets, line 1109: Remove this

lua/diffview/vcs/adapters/git/init.lua
```diff
@@ -0,0 +1,1717 @@
...
+  },
+  ---@type FlagOption[]
+  options = {
+    {
+      "=r", "++rev-range=", "Show only commits in the specified revision range",
+      ---@param panel FHOptionPanel
+      completion = function(panel)
+        return function(arg_lead, _, _)
+          local view = panel.parent.parent
+          return view.adapter:rev_completion(arg_lead, {
```
sindrets, line 1394: `rev_completion` probably needs to be made part of the `VCSAdapter` interface.

lua/diffview/vcs/adapters/git/init.lua
```diff
@@ -0,0 +1,1717 @@
...
+  --     return {}
+  --   end
+
+  --   ---@cast git_toplevel string
+  --   git_dir = vcs.git_dir(git_toplevel)
+  -- end
+
+  -- if not (git_toplevel and git_dir) then
+  --   return {}
+  -- end
```
sindrets, line 1594: Remove this

lua/diffview/vcs/adapters/git/rev.lua
```diff
@@ -0,0 +1,107 @@
...
+    ---@cast revision string
+    self.commit = revision
+  elseif type(revision) == "number" then
+    ---@cast revision number
+    self.stage = revision
+  end
+end
+
+---@param name string
+---@param adapter? GitAdapter
```
sindrets, line 49: This parameter should not be optional.

lua/diffview/vcs/init.lua
```diff
@@ -0,0 +1,44 @@
+local utils = require('diffview.utils')
```
sindrets, line 1: Unused, remove.

lua/diffview/vcs/utils.lua
```diff
@@ -0,0 +1,602 @@
...
+  queue_sync_job(job)
+
+end, 3)
+
+---@param adapter VCSAdapter
+---@param left Rev
+---@param right Rev
+---@param args string[]
+---@param kind git.FileKind
+---@param opt git.utils.LayoutOpt
```
sindrets, line 160: `git.utils.LayoutOpt` this type definition is missing or renamed.

lua/diffview/vcs/utils.lua
```diff
@@ -0,0 +1,602 @@
...
+
+---@param adapter VCSAdapter
+---@param left Rev
+---@param right Rev
+---@param opt git.utils.LayoutOpt
+---@param callback function
+local untracked_files = async.wrap(function(adapter, left, right, opt, callback)
+  Job:new({
+    command = adapter:bin(),
+    args = adapter:get_files_args(),
```
sindrets, line 289: `get_files_args`: if this method's parameter is optional, then it must be declared as such in it's annotations.

lua/diffview/vcs/utils.lua
```diff
@@ -0,0 +1,602 @@
...
+        end)
+        latch:count_down()
+      end
+    end
+  )
+
+  if not (left.type == RevType.STAGE and right.type == RevType.LOCAL) then
+    latch:count_down()
+  else
+    local left_rev = adapter:head_rev() or adapter.Rev.new_null_tree()
```
sindrets, line 393: `head_rev`: This is not defined in the interface

lua/diffview/vcs/utils.lua
```diff
@@ -0,0 +1,602 @@
...
+---Restore a file to the state it was in, in a given commit / rev. If no commit
+---is given, unstaged files are restored to the state in index, and staged files
+---are restored to the state in HEAD. The file will also be written into the
+---object database such that the action can be undone.
+---@param adapter VCSAdapter
+---@param path string
+---@param kind '"staged"'|'"working"'
+---@param commit string
+M.restore_file = async.wrap(function(adapter, path, kind, commit, callback)
+  local undo = adapter:file_restore(path, kind, commit)
```
sindrets, line 450: `file_restore`: This is not defined in the interface.

lua/diffview/scene/views/file_history/option_panel.lua
```diff
@@ -1,7 +1,7 @@
 local lazy = require("diffview.lazy")
 
 local EventEmitter = lazy.access("diffview.events", "EventEmitter") ---@type EventEmitter|LazyModule
```
sindrets, line 3: Remove unused imports.

sindrets (changes requested):
Great work! 

A lot of the comments I made are requesting changes to address diagnostics. I highly recommend that you use a recent version of the [`lua-language-server`](https://github.com/sumneko/lua-language-server), as that is what's driving the typing system here, and the reason why I'm strict about using accurate annotations.

There was one particularly nasty performance regression to the file history rendering, that was caused by - what I assume was - a left over debug line. Easily addressed of course, but take a look at the comment I left in `file_history/render.lua` :)

lua/diffview/scene/views/file_history/render.lua
```diff
@@ -227,6 +227,7 @@ return {
     comp:add_line(cached.root_path)
 
     local offset
+    vim.inspect(panel)
```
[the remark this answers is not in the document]
zegervdv: oops, I cleaned up all `print()` calls, but this one wasn't going to do much good in debugging either ;)

lua/diffview/vcs/adapter.lua
```diff
@@ -0,0 +1,252 @@
...
+
+local M = {}
+
+---@class vcs.adapter.LayoutOpt
+---@field default_layout Diff2
+---@field merge_layout Layout
+
+---@class VCSAdapter: diffview.Object
+---@field bootstrap boolean[]
+---@field context string[]
```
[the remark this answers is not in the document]
zegervdv: something like this: `---@field flags table<string, FlagOption[]>` ?

zegervdv:
Thanks for the thorough review, I think I fixed all of the points, but those I wasn't sure about I left unresolved.

After properly setting up lua-language-server (I probably had a old or broken install) I am still seeing some diagnostic warnings in `vcs/init.lua` at line 36. I'll have a look at that later.

zegervdv:
I think I've cleaned up the last diagnostic issues and expanded the type definitions for the flags and ctx a bit more.

I've updated the implementation for `add_file` and `reset_file` (renamed them to plural, since that makes more sense).

sindrets (commented):
Great! I updated the API for plugins that integrate diffview (currently only Neogit), and I made a few other small changes. I'll have to give the Neogit guys a heads up because we have now changed one of the require paths that they use. Other than that, I reckon this PR is ready! 

I'm gonna test a few more things over the weekend to make sure that everything is in order, just because this is such a big refactor.

zegervdv:
Alright, thanks!

I made a lot of intermediate commits, do you want me to push a single squashed commit or will you do that via the merge?

I ran stylua on the code, but noticed that a lot of files/regions I did not touch also changed. I'm using stylua 0.15.1. You want me to only format the files this PR touches or just leave it as is?

sindrets (changes requested):
> I made a lot of intermediate commits, do you want me to push a single squashed commit or will you do that via the merge?

Don't worry about it; I squash all PR's on merge.

> I ran stylua on the code, but noticed that a lot of files/regions I did not touch also changed. I'm using stylua 0.15.1. You want me to only format the files this PR touches or just leave it as is?

Yeah, I haven't been using stylua in a while. I don't really like any of the lua formatters I've tried. It particularly bothers me how stylua will collapse multiple lines into one when I've explicitly broken it into multiple lines for better readability. And to my knowledge there's no way around this other than to litter your code with `-- stylua: ignore`. And at that point I feel like the effort of circumventing the formatter outweighs its utility.

Sorry about the rant. TLDR: Don't worry about it, just leave it as is :)

---

In my testing over the weekend the only issues I found were with cases of passing invalid options to the commands. We should handle these cases gracefully instead of waiting until the program runs into a runtime error.

lua/diffview/lib.lua
```diff
@@ -25,98 +25,34 @@ function M.diffview_open(args)
...
-  }
-
-  if not git_ctx.dir then
-    utils.err(
-      ("Failed to find the git dir for the repository: %s")
-      :format(utils.str_quote(git_ctx.toplevel))
-    )
-    return
-  end
+  local opts = adapter:diffview_options(args)
```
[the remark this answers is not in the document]
zegervdv: I think it makes sense to add a message here, since diffview_options will not raise an error?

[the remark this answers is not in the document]
sindrets: It *does* echo errors through `GitAdapter:parse_revs()` and `GitAdapter:symmetric_diff_revs()`. I suppose that - API wise - it would be cleaner if these methods returned the error message rather than echoing the error and returning `nil`. But let's not change more in this PR right now.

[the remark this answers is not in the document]
zegervdv: ok, I'll remove the print again.

sindrets (approved):
Alright, let's merge this :)

Thank you for your hard work!

zegervdv:
Thanks for being so supportive! It was a bit daunting at first, doing such a refactor on someone else’s code :). 

I’ve already started implementing the hg adapter so I hope to have a next PR in a couple of weeks.

PR #238, opened 2022-10-24
fix: avoid crash when closing the tab while still updating

raw1z:
When diffview's tab is closed while the commit list is still updating (update can take a few seconds for big repositories), we get an error:

```
Error executing vim.schedule lua callback: ...diffview/scene/views/file_history/file_history_panel.lua:214: attempt to index field 'option_panel' (a nil value)
stack traceback:
        ...diffview/scene/views/file_history/file_history_panel.lua:214: in function <...diffview/scene/views/file_history/file_history_panel.lua:214>
```

### DIFF
--- a/lua/diffview/scene/views/file_history/file_history_panel.lua
+++ b/lua/diffview/scene/views/file_history/file_history_panel.lua
@@ -211,7 +211,11 @@ function FileHistoryPanel:update_entries(callback)
         -- The parent view has closed: shutdown git jobs and clean up.
         finalizer()
         update:close()
-        vim.schedule(function() self.option_panel:sync() end)
+        vim.schedule(function()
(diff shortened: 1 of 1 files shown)


sindrets (approved):
LGTM, thanks!

sindrets:
> (update can take a few seconds for big repositories)

@raw1z *Multiple* seconds? Is that with the default commit limit of 256? What OS is this? For reference, my 6+ year old laptop uses ~500ms to load history for repositories with complex git histories, and ~200ms otherwise.

raw1z:
I'm on macos. I have an intel mac and a m1 mac. I experience the multiple seconds delay only on the m1 mac for the same repo (https://gitlab.com/raw1z/vim)

PR #249, opened 2022-11-17
fix: Regression in finding top-level

FelisNivalis:
`path` should be put in `cwd`, otherwise got "Not a repo" error when the file was in another git repo.


sindrets (approved):
LGTM, thanks!

PR #255, opened 2022-11-23
Implement mercurial (hg) adapter

zegervdv:
Current status:
- [x] DiffviewFileHistory basic list
- [ ] FileHistory add all options
- [x] DiffviewOpen diff view
- [ ] DiffviewOpen merge conflicts
- [ ] documentation

This PR includes a couple of fixes and refactors and adds a new hierarchy to the config (splitting options between hg and git).
I will create a separate PR for these soon, so they can be merged in first with the necessary deprecation logic and migration documentation. Those commits are not the intended focus of this PR.

I already wanted to open this one to ask some questions about the implementation.

As stated above, I have the file history and basic diffview working for mercurial repos, but I'm running into an issue when encountering a merge-conflicted file.
In git there is a single command (created via `GitAdapter:get_namestat_args`) which lists both modified and conflicted files.
For mercurial, these are two separate commands (`hg status -mard` and `hg resolve --list`).
I'm still looking for a command or combination of options that might give me the same result from a single command, but chances are it is not possible.

So, I propose to make a second job with a second args command, e.g. `VCSAdapter:get_unresolved_args`. If this returns `nil`, like will be the case for git, the job will not run and all remains the same as it is now.
If it does return a list of arguments, it will run and the results of both jobs (namestat and unresolved) is merged and parsed as if it was a single command.
Does that sound like a reasonable approach or do you know a better/cleaner way?


sindrets:
I think it would make more sense to move the implementation of [`tracked_files()`](https://github.com/zegervdv/diffview.nvim/blob/f953f9bb274da33d440cdb7cd2a19efeedae0dd1/lua/diffview/vcs/utils.lua#L162) and [`untracked_files()`](https://github.com/zegervdv/diffview.nvim/blob/f953f9bb274da33d440cdb7cd2a19efeedae0dd1/lua/diffview/vcs/utils.lua#L264) into the adapters. It's natural that these would have a slightly different implementation for different vcs. This way you can use any number of jobs you need for mercurial (although it should be kept to a minimum to reduce I/O overhead).

zegervdv:
@sindrets 
I think I have most things working now, maybe just need to add some completions and some of the lesser used flags.

I'll test this for a while to cover all the corner cases.

I noticed that hg is quite a bit slower in generating the file history information: on a repo of ~50K commits it takes about 15 to 20 seconds to list 256 commits. Since hg is a python script, each invocation to the command takes a performance hit when it starts the python interpreter. There are some options to improve it: for example there is a experimental `chg` command which starts up much faster (which gets the file history down to about 5 seconds for the previously mentioned testcase).
Another option is to start hg as a command server (so it starts only once) and send commands to it. Implementing this seems a much bigger change :).

For maximum compatibility, I will keep the default `config.hg_cmd` as `hg`, but mention in the docs that a speed up can be achieved by substituting it with `chg`.

How would you prefer handling the docs?
Currently most of it is written with git commands in mind, which will probably still be 99.8% of the users after this is merged :).
So with that in mind I think a separate section is probably the better approach?

sindrets:
@zegervdv Great work!

> on a repo of ~50K commits it takes about 15 to 20 seconds to list 256 commits

Ouch, that's pretty bad 😅. Does the history size matter for the performance when you're only querying `n` commits? Are you passing the appropriate flags to limit the output? If this can't easily be remedied I think it's best to quite drastically lower the default commit limit for hg. While we *do* make sure to kill jobs if a history view is closed during an update, and we *do* make sure that jobs are non-blocking; lua coroutines are still single threaded, and the editor isn't exactly super responsive during file-history updates. To have the editor be like that for almost half a minute is far from ideal.

> Another option is to start hg as a command server (so it starts only once) and send commands to it. Implementing this seems a much bigger change :).

This sounds interesting. If hg has a daemon like you describe, it sounds like the appropriate way to use it in a third party tool like this where we are making up to hundreds of calls to the VCS. But I also agree that it shouldn't be a blocker to get this PR merged. Just something to explore for the future.

> How would you prefer handling the docs?

For the commands, we can just have one section of options for each VCS tool. I'll push a commit where I've started to amend the docs to show what I mean, and then you can just fill it out with descriptions of the different options you've implemented :)

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1051 @@
...
+        additions = tonumber(adds),
+        deletions = tonumber(changes) - tonumber(adds),
+      }
+    end
+
+    if not (kind == "staged") then
+      file_info[name] = {
+        status = status,
+        name = name,
+        oldname = name, -- TODO
```
sindrets, line 781: This field should only be defined if the file in question really has been renamed and `oldname` is different from `name`. Other parts of the code will default to showing the `oldname` if it's present, and assume that the file has been renamed.

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1051 @@
...
+    end
+  end
+
+  local find_key = function (t, key, value)
+    for _, v in ipairs(t) do
+      if v[key] == value then
+        return v
+      end
+    end
+  end
```
sindrets, line 793: Instead of iterating over the json data every time we're looking for a value, it would be a lot better to process the json data once, producing a struct that let's you access what you need directly.

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1051 @@
...
+  -- silently.
+  -- Solution: queue them and run them one after another.
+  vcs_utils.queue_sync_job(job)
+end, 4)
+
+HgAdapter.flags = {
+  ---@type FlagOption[]
+  switches = {
+    {'--follow', '-f', 'Follow renames'},
+    {'--no-merges', '-M', 'List no merge changesets'},
```
sindrets, line 957: The order should be `keymap`, `flag_name`, `description`. So:

```suggestion
    { '-f', '--follow', 'Follow renames' },
    { '-M', '--no-merges', 'List no merge changesets' },
```

sindrets (changes requested):

sindrets:
@zegervdv If you would like to, I'll make you a collaborator on the project after we've merged this. That way you can more effectively maintain the hg part of the code when you have time :)

zegervdv:
I did a couple of tests on different repos and it seems the total number of commits in the repo is not a (huge) factor.

The limitation to 256 works, I can see the counter increment as they come in and setting it to a lower value also confirms that I get the right number of lines.
So perhaps reducing it to 64 as a default makes sense.

>If hg has a daemon like you describe, it sounds like the appropriate way to use it in a third party tool

Yes, the vscode plugin for mercurial has it as an optional feature, so I'm not sure how easy/stable it is.
But it could be a fun exercise for later.

Definitively if the new sapling-scm also has it and would become a new major player. 


Thanks for the reviews, there are so many features in this plugin I haven't even discovered them all, let alone had a chance to use them :) 


> @zegervdv If you would like to, I'll make you a collaborator on the project after we've merged this. That way you can more effectively maintain the hg part of the code when you have time :)

That would be nice, thanks for the trust :)

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+
+  local options = {
+    show_untracked = true, -- TODO: extract from hg config
+    selected_file = argo:get_flag("selected-file", { no_empty = true, expand = true })
+      or (vim.bo.buftype == "" and pl:vim_expand("%:p"))
+      or nil,
+  }
+
+  return {left = left, right = right, options = options}
+end
```
sindrets, line 678: The revs here are hardcoded and do not respect the rev argument passed by the user. The rev arg should be the first non-option argument, which can be accessed like this: `argo.args[1]`. Should probably implement something like the `GitAdapter:parse_revs()` method.

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+
+function VCSAdapter:rev_to_pretty_string(left, right)
```
sindrets, line 680:
```suggestion
function HgAdapter:rev_to_pretty_string(left, right)
```

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+    return left:abbrev() .. "::" .. right:abbrev()
+  end
+  return nil
+end
+
+function HgAdapter:head_rev()
+  local out, code = self:exec_sync({ "log", "--template={node}", "--limit=1", "--"}, {cwd = self.ctx.toplevel, retry_on_empty = 2})
+  if code ~= 0 then
+    return
+  end
```
sindrets, line 695:
```suggestion
  local out, code = self:exec_sync({ "log", "--template={node}", "--limit=1", "--"}, {
    cwd = self.ctx.toplevel,
    retry_on_empty = 2,
  })

  if code ~= 0 then
    return
  end
```

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+  if kind == "working" and next(conflict_map) then
+    for _, v in pairs(conflict_map) do
+      table.insert(conflicts, FileEntry.with_layout(opt.merge_layout, {
+        adapter = self,
+        path = v.name,
+        oldpath = v.oldname,
+        status = "U",
+        kind = "conflicting",
+        revs = {
+          a = self.Rev(RevType.COMMIT, nodes['local']),
```
sindrets, line 917:
```suggestion
          a = self.Rev(RevType.COMMIT, nodes.local),
```

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+    { '=r', '--rev=', 'Revspec', prompt_label = "(Revspec)" },
+    { '=l', '--limit=', 'Limit the number of changesets' },
+    { '=u', '--user=', 'Filter on user' },
+    { '=k', '--keyword=', 'Filter by keyword' },
+    { '=b', '--branch=', 'Filter by branch' },
+    { '=B', '--bookmark=', 'Filter by bookmark' },
+    { '=I', '--include=', 'Include files' },
+    { '=E', '--exclude=', 'Exclude files' },
+  },
+}
```
sindrets, line 1060: I just recently changed how these flag options are declared, sorry about that 😅 Take a look in the git adapter:
https://github.com/zegervdv/diffview.nvim/blob/a281b78204a5b3d2b54ff395df094824fac7b09e/lua/diffview/vcs/adapters/git/init.lua#L1656

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+          end, value), " ")
+        end
+        return utils.str_quote(value, { only_if_whitespace = true })
+      end,
+    })
+
+    list[i] = option
+    list[option.key] = option
+  end
+end
```
sindrets, line 1105: No longer needed. Remove.

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+
+function HgAdapter:is_binary(path, rev)
+  -- TODO
+  return false 
+end
+
+-- TODO: implement completion
+function HgAdapter:rev_completion(arg_lead, opt)
```
sindrets, line 1113:
```suggestion
function HgAdapter:rev_candidates(arg_lead, opt)
```

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+  return { }
+end
+
+function HgAdapter:init_completion()
+  self.comp.file_history:put({"--rev", "-r"}, function(_, arg_lead)
+    return self:rev_completion(arg_lead, { accept_range = true })
```
sindrets, line 1119:
```suggestion
    return self:rev_candidates(arg_lead, { accept_range = true })
```

sindrets (changes requested):

lua/diffview/vcs/adapters/hg/rev.lua
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  return "UNKNOWN"
+end
+
+function HgRev.to_range(rev_from, rev_to)
+  if rev_to then
+    return rev_from .. "::" .. rev_to
+  end
+  return rev_from
```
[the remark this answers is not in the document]
zegervdv: In hg we use a revset which can be one or more commits:

-  `hg log -r <rev>`: will give the log of that commit only
- `hg log -r ::<rev>`: gives all commits from start (commit 0) to rev
- `hg log -r <rev>::`: gives all commits from rev until last commit
Also, there is a difference between (`<rev a>::<rev b>` and `<rev a>:<rev b>`, the latter will give all commits even on different branches, the first only on the current branch, where branch is a sequence of commits, not necessarily a named branch)

If you do `hg diff -r <rev>`, you will get the diff between `<rev>` and the current revision. `hg diff -r <rev a>::<rev b>` will give the diff between the two revisions.

Since there is in most cases no implicit range, I thought it would be better to leave it to the user?

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+      ret.ours = { hash = commit.node }
+      out, code = self:exec_sync({ "log", "--template={branch}", "--rev", commit.node })
+      if code == 0 then
+        ret.ours.ref_name = out[1]
+      end
+    end
+  end
+
+  -- Base is ancestorlinknode (per file)
+  ret.base = { hash = "" }
```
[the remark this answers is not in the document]
zegervdv: It seems hg gives me the merge-base per file, which is not always the same.
So to be able to give that commit, I would need to know which file is being viewed currently. Maybe something for a follow up?

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1141 @@
...
+    { "exclude", "X" },
+  }
+
+  ---@type LogOptions
+  local log_options = { rev_range = range_arg }
+  for _, names in ipairs(log_flag_names) do
+    local key, _ = names[1]:gsub("%-", "_")
+    local v = argo:get_flag(names, {
+      expect_string = type(config.log_option_defaults[self.config_key][key]) ~= "boolean",
+      expect_list = names[1] == "L",
```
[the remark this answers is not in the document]
zegervdv: yes, I haven't found an equivalent (sadly, seems like a useful feature, kinda replaces the need for `git blame`)

lua/diffview/vcs/adapters/hg/init.lua
zegervdv: In hg the rev is always passed via `-r <revset>` (or `--rev=<revset>`), so not as the unnamed argument like it is in git.

In all other cases we use the same command options and flags, so I thought it makes sense to do that here too?

lua/diffview/vcs/adapters/hg/init.lua
zegervdv: `local` is a keyword, so it doesn't allow it as a name :)

lua/diffview/vcs/adapters/hg/init.lua
zegervdv: For this function, I'm not sure how to implement.
With the revsets you get some kind of [mini-language](https://hg.mozilla.org/mozilla-central/help/revsets) for filtering, I don't want to write a whole parser for that :sweat_smile:

lua/diffview/vcs/adapters/hg/init.lua
zegervdv: No worries, looks like a straightforward change

lua/diffview/vcs/adapters/hg/init.lua
sindrets: It's not as clear cut as that. `:DiffviewOpen` isn't directly synonymous with `git diff`. It was designed to be similar to the git command of course, but it has its own command structure. For instance the git command allows you to specify the range as two separate rev args, our command does not. The git command generally allows you to specify paths without the end-of-options token (`--`) as long as the args are not ambiguous. Our command does not (to avoid dealing with arg ambiguity). And finally, our command has some custom options that are not from git. I kind of regret not naming these options differently such that it would be clear that they are not git options, but oh well, that's off-topic.

Point is, `:DiffviewOpen` is it's own thing, and it has its own command structure that I would like to stick to:

```
:DiffviewOpen [rev] [options] [-- [paths...]]
```

lua/diffview/vcs/adapters/hg/init.lua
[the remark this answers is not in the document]
sindrets: It seems like it was added in hg recently actually. It's a documented option for `hg log` on my install at least:

```
       -L,--line-range <FILE,RANGE[+]>
              follow line range of specified file (EXPERIMENTAL)
```

But we can look at this later :)

lua/diffview/vcs/adapters/hg/init.lua
sindrets: Ah, right. My bad 😅

lua/diffview/vcs/adapters/hg/rev.lua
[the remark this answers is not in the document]
sindrets: > - `hg log -r <rev>`: will give the log of that commit only
> - `hg diff -r <rev>`, you will get the diff between `<rev>` and the current revision

Right, it's this type of disparity that `to_range()` is supposed to circumvent. Would it work to do something like this maybe:

```
<rev>^::<rev>
```

lua/diffview/vcs/adapters/hg/init.lua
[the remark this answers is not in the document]
sindrets: Really? The merge base can be different for different files within a single conflict? How does that work? Are we confusing terminology here perhaps? In git the merge base refers to the common ancestor of all the branches involved in a merge. Thus, by definition, there can only be one.

But sure, we can address this in a follow up PR. Add a `TODO` comment here or something so that we don't forget to come back to this :)

lua/diffview/vcs/adapters/hg/init.lua
sindrets: Wow, that's really cool! But, yeah, not need to write a complex parser for that. Completion is only a helpful tool that aims to reduce the friction of typing out a command by hand. Simply listing all branches for instance seems like it would be helpful. And it seems simple to support the most common (?) types of ranges as well, like:

- `<rev>::<rev>`
- `<rev>:<rev>`

The code for this would be really similar to what's currently in the git adapter.

lua/diffview/vcs/adapters/hg/init.lua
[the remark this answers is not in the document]
zegervdv: I created a little test case with 2 files, and I do get a merge-base revision per file (it's called `ancestorlinknode`) but it is indeed the same for all files involved, except if the file is new (then it is `00000000`).

The testcase I was using before was a bit of a stretch, I had created the same new file in both branches.

So I guess I can use the first non-zero `ancestorlinknode` and assume it is the same for all.

lua/diffview/vcs/adapters/hg/init.lua
zegervdv: Yes, this and the [changeset evolution](https://www.mercurial-scm.org/doc/evolution/) are two things I miss when working with git :) 

I'll add those basic range constructs indeed

lua/diffview/vcs/adapters/hg/rev.lua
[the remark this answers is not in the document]
zegervdv: Oh, I think I understand now: you always want a range, even though it is a single change.

lua/diffview/vcs/adapters/hg/init.lua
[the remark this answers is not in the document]
zegervdv: I hadn't seen that (it hidden behind a `--verbose`).
I'll add this in the next PR :)

lua/diffview/vcs/adapters/hg/init.lua
zegervdv: Alright, I'll fix it

lua/diffview/vcs/adapters/hg/rev.lua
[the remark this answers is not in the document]
zegervdv: In hg I use `hg diff -c <rev>` to the the change this commit introduces but that is just a shorthand for `<rev>^::<rev>` indeed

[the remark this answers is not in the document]
zegervdv: No, I was wrong.
`hg diff -r` is not documented. It works as described above, but `hg log` does not.

Seems like I need to use `--from <rev> --to <rev>` for `diff`.

I'll need to do some more tests to make sure this is always giving me the right data.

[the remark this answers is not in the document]
zegervdv: It took me a bit longer than I'd like to admit, but I think I figured it out.
@sindrets: just to confirm this is the right behaviour:
  * `DiffviewOpen` shows the uncommitted changes (there is no staging in hg)
  * `DiffviewOpen 3` shows the diff between revision 3 up to and including uncommitted changes
  * `DiffviewOpen 3::5` shows the diff between revision 3 up to and including  5 (i.e., changes introduced in 4 and 5)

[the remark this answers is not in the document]
sindrets: Yes, that's correct.

lua/diffview/vcs/adapters/hg/init.lua
```diff
@@ -1,18 +1,1258 @@
...
+end
+
+function HgAdapter:rev_to_pretty_string(left, right)
+  if left.type == RevType.CUSTOM then
+    return left.commit
+  elseif right and right.commit then
+    return left:abbrev() .. "::" .. right:abbrev()
+  end
+  return nil
+end
```
sindrets, line 736: Did you mean to use `RevType.CUSTOM` here? If so, why? This rev type is a remnant of `:h diffview-api` that I don't think anyone has ever used, and I'm probably going to remove it in the future.

Also, why is the logic in this method different from the logic in the super method?

lua/diffview/vcs/adapters/hg/rev.lua
```diff
@@ -0,0 +1,58 @@
...
+  t = type(track_head)
+  assert(t == "boolean" or t == "nil", "'track_head' must be of type boolean!")
+
+  self.type = rev_type
+  self.track_head = track_head or false
+
+  self.commit = revision
+end
+
+function HgRev:new_null_tree()
```
sindrets, line 32: This should be a static method:

```suggestion
function HgRev.new_null_tree()
```

sindrets (changes requested):
https://github.com/zegervdv/diffview.nvim/blob/673b58fe1a0b13f4a4fdcca5166ac41a2156c5c9/lua/diffview/scene/views/diff/diff_view.lua#L128

This line should probably have an added condition of:

```lua
if --[[ ... ]] and self.adapter:instanceof(GitAdapter.__get()) then
```

---

This PR seems almost good to go! If you address the few things I mentioned here, I think I'd be ready to merge :)

lua/diffview/vcs/adapters/hg/init.lua
zegervdv: right, it's a leftover from some trials

zegervdv:
Fixed the last comments. I think it is ready :)

sindrets (approved):
LGTM, let's get this merged!

sindrets:
@zegervdv I've also sent you an invite to become a collaborator on the project :)

zegervdv:
Thanks!

PR #257, opened 2022-11-24
feat(help): Help panel popup

### LINKED ISSUE #168: [Feature] action to view current keybinds
There is quite a lot of mappings that are possible and, unfortunately, I have a terrible memory.

It could be nice to have an action that would show all the current keybinds for the current context (depending on whether we are in the view, file_panel, file_history_panel, etc). Something like `actions.show_keymaps`

<details>
<summary><b>Examples:</b></summary>

For example, with Neotree pressing `?` will show 

image

Or similar thing in Undotree:

image

</details>

JMarkin:
Resolves #168.

Hello, I tried add desc and popup menu for keymaps.


sindrets:
Thanks for the PR!

I see that you're using the Nui library here. I would like to avoid relying on additional dependencies. Especially for something as simple as creating a buffer in a floating window. That, and we already have [a Panel abstraction](https://github.com/sindrets/diffview.nvim/blob/main/lua/diffview/ui/panel.lua) that allows us to do this quite easily. For an example, see i.e. [the commit log panel](https://github.com/sindrets/diffview.nvim/blob/main/lua/diffview/ui/panels/commit_log_panel.lua).

The way you have changed the structure of the keymap declerations in the config will unnecessarily break compatibility with vim command maps.

Also, please do not format parts of the code that you didn't touch. It makes it harder to review the changes.

Let me know if you have any questions about implementation :)

JMarkin:
@sindrets Thanks for your review. I rewrote by your comment. 

I tried to create runnable command in list, but I didn't. Before ready for merge I want to complete this feature. Can you give me a tip how to run command for current panel(not help panel), if you know ? If you think this feature is overrate, I will remove it.

sindrets:
> I tried to create runnable command in list, but I didn't.

I don't understand what that is. Can you please elaborate?

JMarkin:
@sindrets It's like. I have menu -> press "enter" -> command under cursor would be apply.
image

JMarkin:
I did it, Please review)

sindrets (approved):
Thank you for your contribution! I refactored it a little bit.

danielo515:
Awesome! This is fantastic

danielo515:
This is awesome, but will it be possible to add a little annotation at the top of the window about how to open the help pane? I think it is not obvious for everybody.

Issue #241, opened 2022-11-01
Question: Different highlight in both panels


jgarciaokode:
Hi, first of all, thanks for this awesome plugin

I am wondering how do you manage to get this highlight? 

<img width="1635" alt="image" src="https://user-images.githubusercontent.com/79897391/199238194-f4368ec1-0923-4275-b336-4247296b6fec.png">

I mean i just manage to make diffchange with a background and difftext with another background but both left and right side panels have the same background from diffchange. In your image you have left side red right side green and in both panels it seems like you highlight difftext differently

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/79897391/199238342-5bf3f008-3af9-460a-81ed-c7becbb84a6a.png">

sindrets:
@jgarciaokode I have a rather involved setup in my own config where I [dynamically generate diff highlights](https://github.com/sindrets/dotfiles/blob/cdaae8429e1a041073e9d3e7ea318bfb8a072ce7/.config/nvim/lua/user/colorscheme.lua#L148). Then I use some undocumented parts of the internal plugin API to [adjust `'winhighlight'` for the two diff windows](https://github.com/sindrets/dotfiles/blob/cdaae8429e1a041073e9d3e7ea318bfb8a072ce7/.config/nvim/lua/user/plugins/diffview.lua#L61-L83).

This is not builtin functionality in the plugin because the highlight generation makes some assumptions about how diff colors should be presented that would be very difficult to generalize such that it would produce good results for any arbitrary color scheme. There's simply too much variation in how color scheme creators choose to approach diff highlights.

jgarciaokode:
@sindrets Many thanks for the detailed explanation, i will have a look at your config and give it a try.

jgarciaokode:
In case someone is interested, i managed to get it working pretty easily with this modifications.

Step 1: Set this highlights groups

`vim.cmd([[highlight DiffAdd gui=none guifg=none guibg=#103235]])
vim.cmd([[highlight DiffChange gui=none guifg=none guibg=#272D43]])
vim.cmd([[highlight DiffText gui=none guifg=none guibg=#394b70]])
vim.cmd([[highlight DiffDelete gui=none guifg=none guibg=#3F2D3D]])
vim.cmd([[highlight DiffviewDiffAddAsDelete guibg=#3f2d3d gui=none guifg=none]])
vim.cmd([[highlight DiffviewDiffDelete gui=none guifg=#3B4252 guibg=none]])

[...]

PR #258, opened 2022-11-25
feat: Allow enhanced diff highlight customization

antosha417:
Hey! I really enjoy using this plugin and like how `enhanced_diff_hl` looks. The idea that we can apply different highlight groups to different windows blew my mind. 
So I want to take it even further and allow user customization by allowing `enhanced_diff_hl` be an object. 

What do you think about it? 

I'll add this feature to the docs if you would like to merge it. 
 
The config will look like this:
```
require("diffview").setup({
  enhanced_diff_hl = {
    a = {
      "DiffAdd:DiffviewDiffAddAsDelete",
      "DiffDelete:DiffviewDiffDelete",
      "DiffChange:DiffviewDiffChangeAsDelete",
      "DiffText:DiffviewDiffTextRed",
    },
    b = {
      "DiffDelete:DiffviewDiffDelete",
      "DiffChange:DiffviewDiffChangeAsAdd",
      "DiffText:DiffviewDiffTextGreen",
    }
  }
 })
```

### DIFF
--- a/lua/diffview/scene/views/standard/standard_view.lua
+++ b/lua/diffview/scene/views/standard/standard_view.lua
@@ -81,14 +81,20 @@ function StandardView:init_layout()
 end
 
 function StandardView:post_layout()
-  if config.get_config().enhanced_diff_hl then
-    self.winopts.diff2.a.winhl = {
-      "DiffAdd:DiffviewDiffAddAsDelete",
-      "DiffDelete:DiffviewDiffDelete",
-    }
-    self.winopts.diff2.b.winhl = {
-      "DiffDelete:DiffviewDiffDelete",
-    }
+  local enhanced_diff_hl = config.get_config().enhanced_diff_hl
+  if enhanced_diff_hl then
+    if (type(enhanced_diff_hl) == "table") then
+      self.winopts.diff2.a.winhl = enhanced_diff_hl.a
+      self.winopts.diff2.b.winhl = enhanced_diff_hl.b
+    else
+      self.winopts.diff2.a.winhl = {
+        "DiffAdd:DiffviewDiffAddAsDelete",
+        "DiffDelete:DiffviewDiffDelete",
+      }
+      self.winopts.diff2.b.winhl = {
+        "DiffDelete:DiffviewDiffDelete",
+      }
+    end
   end
 end
 


sindrets:
Thanks for the PR!

I've been considering allowing more customization for the `enhanced_diff_hl`, but I'm not sure this is the right approach. Your suggested implementation would kind of lock us into only offering enhanced highlights for the 2-way split layouts. I'm gonna take some time to think about what the best approach would be.

I would like to also point out that you can actually already do this by using the hooks. Take a look at how I'm customizing the highlights [in my own dotfiles](https://github.com/sindrets/dotfiles/blob/a3f470f7b4b702543d91cdd2d4e301bef273a6d1/.config/nvim/lua/user/plugins/diffview.lua#L61-L83).

I'll also link an issue where someone asked about something similar: #241

antosha417:
Thank you! I managed to get it working using hooks. 
Didn't understand the part `view.emitter:on("post_layout", post_layout)` though.
I think it should be easier to configure the colors without getting into internal parts of the plugin.

In this pr I suggested to accept a `table` instead of `boolean` in `enhanced_diff_hl` setting. Now after I copied this hook from you I think that we can instead accept a `function`. That function would return an object like. 
```lua
{
  a = {
    winhl = {
        "DiffAdd:DiffviewDiffAddAsDelete",
        "DiffDelete:DiffviewDiffDelete",
      },
  b = {...}, 
  c = {...},
  ...
}
```

What do you think about this approach?

sindrets:
Sorry for the late response! I've concluded that the use-case for this is too specific to warrant its own config option. I've opted to instead provide more context data to the hooks, which are much better suited for this sort of customization, as they allow for dynamic changes and don't rigidly limit the type of customization.

As an example, adjusting the `'winhl'` only for `diff2*` layouts can now be done like this:

```lua
require("diffview").setup({
  hooks = {
    diff_buf_win_enter = function(bufnr, winid, ctx)
      if ctx.layout_name:match("^diff2") then
        if ctx.symbol == "a" then
          vim.opt_local.winhl = table.concat({
            "DiffAdd:DiffviewDiffAddAsDelete",
            "DiffDelete:DiffviewDiffDelete",
          }, ",")
        elseif ctx.symbol == "b" then
          vim.opt_local.winhl = table.concat({
            "DiffDelete:DiffviewDiffDelete",
          }, ",")
        end
      end
    end,
  }
})
```

See the updated docs for more info about the hook callback parameters:

```
:h diffview-config-hooks
```

antosha417:
Thank you for the new approach! Will give it a try tomorrow

tummetott:
I really like the approach, how you can style individual windows inside a diffview. One question remains:

When filtering for 3 way diffs (in fact a merge conflict) with the following code, the `ctx.symbol` is empty.
```lua
if ctx.layout_name:match("^diff3") then
    vim.pretty_print(ctx)
end
```
the output is:
```lua
{
  layout_name = "diff3_horizontal"
}
{
  layout_name = "diff3_horizontal"
}
{
  layout_name = "diff3_horizontal"
}
```
Is this intended behaviour? I'd also love to style my colors for a merge conflict

sindrets:
@tummetott Thank you for reporting! No, that wasn't intended behavior. Should be fixed now.

PR #260, opened 2022-11-30
feat: Stage individual hunks by editing index buffers

sindrets:
This finally brings over a really nice feature from Fugitive that allows you to stage individual hunks by editing any buffer that represents the index (after running `:DiffviewOpen` with no `[git-rev]` the entries under "Changes" will have the index buffer on the left side, and the entries under "Staged changes" will have it on the right side). Once you write to an index buffer, the index will be updated.

### For those unfamiliar with the feature from Fugitive:

This makes the staging workflow play really nicely with the builtin `:h copy-diffs` commands. After moving the cursor to a diff hunk in an index buffer, you can simply press `do` to obtain the hunk from the other side. When you now `:w` the index buffer you will have staged that individual hunk.

It's now equally trivial to undo individual hunks from the staged changes. All the entries in the "Staged changes" section show a diff between the HEAD on the left and the index on the right. Thus, bringing the cursor to a hunk in the index buffer, pressing `do` to obtain the content from the HEAD version, and finally `:write`-ing the index buffer will unstage that hunk.

Having editable index buffers gives you great control over exactly what gets staged, as you can now edit the index the same way you edit all text in (N)vim.

### How do I identify an index buffer?

Again, running `:DiffviewOpen` with no `[git-rev]` will have the index buffer on the left side for all the entries under the "Changes" heading, and on the right side for all entries under the "Staged changes" heading.

You can also identify an index buffer from the buffer name. All Diffview buffer names are URI's structured like this:

```
diffview://{git_dir}/{rev}/{path}
```

Where the variables represent the following:

- `git_dir`: Path to the git dir (typicaly the `.git` dir. But not necessarily)
- `rev`: The abbreviated revision object name. Will either be an abbreviated commit SHA, or a stage number `:[0-3]:`.
- `path`: The path to the file from the top-level of the repo.

The index in git is represented by stage number 0. So an example index buffer could be named something like this:

```
diffview:///path/to/git/dir/:0:/path/to/file
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,14 @@ Additional commands for convenience:
 With a Diffview open and the default key bindings, you can cycle through changed
 files with `<tab>` and `<s-tab>` (see configuration to change the key bindings).
 
+#### Staging
+
+You can stage individual hunks by editing any buffer that represents the index
+(after running `:DiffviewOpen` with no `[git-rev]` the entries under "Changes"
+will have the index buffer on the left side, and the entries under "Staged
+changes" will have it on the right side). Once you write to an index buffer the
+index will be updated.
+
 ### `:[range]DiffviewFileHistory [paths] [options]`
 
 Opens a new file history view that lists all commits that affected the given
--- a/doc/diffview.txt
+++ b/doc/diffview.txt
@@ -55,6 +55,14 @@ COMMANDS                                        *diffview-commands*
             :DiffviewOpen HEAD~2 -- lua/diffview plugin
             :DiffviewOpen d4a7b0d -uno
 <
+
+                                                *diffview-staging*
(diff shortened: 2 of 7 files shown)


MagicDuck:
This is magical, you should publicize it more in the README.md! 😄 
I just played a little with it and am planning to create mappings for myself in the index buffers for stuff like:
- stage next hunk: `normal! ]cdo`,
- etc.

ltuijnder:
Thank you for this feature! Really nice

Could this feature get documented better? Personally, I was confused when reading the official documentation what is the index file? And I was not aware of the `do`-command (which is also not listed in the `g?`)

The explanation in this PR helped enormously! :) And I would not have minded a more verbose explanation like this pr in the official documentation.

Still thank you for adding this.

PR #271, opened 2022-12-15
Split configuration to allow for other VCS types

zegervdv:
These are some required changes to allow the introduction of other VCS types (see #255 ).
It splits the `file_history_panel.log_options` config table into separate sub-tables per vcs type.

Also includes a couple of refactors to allow a smoother integration.


zegervdv:
In the last commit I fixed a little bug in the logging: it is not allowed to call a `vim.api` or `vim.fn` function from a callback.

sindrets (approved):
@zegervdv LGTM, thanks!

> In the last commit I fixed a little bug in the logging: it is not allowed to call a `vim.api` or `vim.fn` function from a callback.

Thanks for catching this! To expand a little on why / when this happens: this only happens in async lua code. So, all libuv callbacks and inside coroutines. You can work around this by using `:h vim.schedule()`. This will schedule the given function to be run the next time it is safe to call the API.

PR #276, opened 2022-12-21
refactor: move show log command to VCSAdapter

zegervdv:
Seems like I missed a spot :)


sindrets (approved):
@zegervdv Thanks!

Just a heads up: I added a new method to the Rev class that you're gonna have to implement for `HgRev`.

PR #278, opened 2022-12-30
feat: Allow hiding mappings from the help panel

fjchen7:


sindrets:
@fjchen7 Thanks for the PR!

I'm missing some explanation for this change. What's the point of hiding entries from the help panel?

fjchen7:
The final purpose is to make help  panel cleaner if user wants.

For me the specific requirement is:

```lua
require("diffview").setup {
  view = {
    file_panel = {
      -- I don't want to show some conventional keymaps
      { "n", "j", actions.next_entry, { desc = false } },
      { "n", "<down>", actions.next_entry, { desc = false } },
      { "n", "k", actions.prev_entry, { desc = false } },
      { "n", "<up>", actions.prev_entry, { desc = false } },

      -- I want to disable some keymaps withouts showing them in help panel.
      { "n", "<C-o>", "", { desc = false } },  -- Avoid accidentally messing up buffer
      { "n", "<C-i>", "", { desc = false } },
    }
  }
}
```

lua/diffview/ui/panels/help_panel.lua
```diff
@@ -212,10 +215,12 @@ function HelpPanel:render()
     for _, item in ipairs(section.items) do
       ---@cast item CompStruct
       comp = item.comp
+      if comp.context.label_rhs == "diffview_ignored" then goto continue end
```
sindrets, line 218: Don't use `goto` here. Just use normal branching:

```
if comp.context.label_rhs ~= "diffview_ignored" then --[[ ... ]]
```

sindrets (changes requested):
Setting the `desc` field to anything other than a string is unintended usage of the `nvim_set_keymap()` API. While it seems to have no side effects right now; there's no guarantee it will stay that way in the future. We need to do this properly and special-case this behavior in the plugin's keymap processing.

You can check [here](https://github.com/sindrets/diffview.nvim/blob/e3e15d8f1d727056185fc1696be1554837dec897/lua/diffview/config.lua#L403-L411) whether the `desc` field has been set to `false`, and if so, change the value of the expanded mapping to `__DIFFVIEW_IGNORED__` or something.

fjchen7:
Of we directly choose the special string like `__DIFFVIEW_IGNORED__` or something to hide keymap? (Just like `which_key_ignore` in [which-key](https://github.com/folke/which-key.nvim)) For example,

```lua
{ "n", "<2-LeftMouse>", actions.select_entry, { desc = "__DIFFVIEW_IGNORED__" } },
```

In my opinion, this could avoid conflict with the extension about `desc = false` in `nvim_set_keymap()` and may be better.

sindrets:
@fjchen7 good call. Let's just go with `"diffview_ignore"` as the special value then.

sindrets (approved):
Thanks!

PR #290, opened 2023-01-19
conflicts: handle FETCH_HEAD conflicts

emmanueltouzery:
needed this to have diffview correctly appear after I got a merge conflict. `their_head` was nil, as it turns out, because diffview didn't expect that exact scenario.

i'm otherwise not exactly sure what I did in this case that was so special.

### DIFF
--- a/lua/diffview/vcs/adapters/git/init.lua
+++ b/lua/diffview/vcs/adapters/git/init.lua
@@ -257,7 +257,7 @@ end
 function GitAdapter:get_merge_context()
   local their_head
 
-  for _, name in ipairs({ "MERGE_HEAD", "REBASE_HEAD", "REVERT_HEAD" }) do
+  for _, name in ipairs({ "MERGE_HEAD", "REBASE_HEAD", "REVERT_HEAD", "FETCH_HEAD" }) do
     if pl:readable(pl:join(self.ctx.dir, name)) then
       their_head = name
       break


sindrets:
Unfortunately, to my knowledge, git provides no straight forward way to find the commit of THEIR side in a conflict. Thus, I'm left to guess all the ways git might represent that information during a conflict.

However, it seems strange to me that git would use the `FETCH_HEAD` to for this purpose. It would be helpful if you could explain how you initiated that conflict. Surely not by running `git fetch`?

emmanueltouzery:
i believe it might have been...

git fetch upstream
git rebase origin/master

but i'm not 100%. the point is that my local master branch was behind. so i had to rebase based on the remote branch.
maybe there's a difference if it's a remote vs a local branch?

i'll try to double-check through the week-end if you don't beat me to it.

emmanueltouzery:
no, I tried that scenario now and I got... well i have the following files in the .git folder:
```
REBASE_HEAD
MERGE_MSG
AUTO_MERGE
HEAD
ORIG_HEAD
FETCH_HEAD
COMMIT_EDITMSG
```

while if I abort the rebase, I have:
```
HEAD
ORIG_HEAD
FETCH_HEAD
COMMIT_EDITMSG
```

so I don't know. All I know for sure is that diffview crashed because `their_head` was null. and that adding `FETCH_HEAD` in that list made it work for me.

If I get the issue again, i'll try to remember exactly which commands I ran, and maybe what were the contents of the .git folders..

emmanueltouzery:
So the title of the bug is possibly wrong. The real cause is probably something else, bottom line is that making that change fixed a crash for me, but I don't know the real cause

sindrets:
This change only circumvents an assertion error for an unknown condition. And it's probably not even the correct thing to do in that condition. From all the documentation I've read, and my testing, git does not use the `FETCH_HEAD` to represent `THEIR` side during a conflict. So I won't merge this without more information.

If you encounter the bug again, please provide more information like:

- What git commands you ran to initiate the conflict
- The stack trace and error message
- Relevant info from `:DiffviewLog`

emmanueltouzery:
probably makes sense to close this, i'll open a new PR/issue with more info if I get the issue again.

PR #317, opened 2023-03-11
feat(actions): Resolve conflicts for the whole file

616b2f:
Hi there :)

I asked for that functionality and get the advice I should create a PR, so here it comes.

This is my first PR here and also I don't have experience in neovim plugin development, so any tips are welcome and appreciated.

I would also like an advice if I should keep it in a separate funktion "conflict_choose_file()" or merge it into "conflict_choose()", what do you think?


lua/diffview/config.lua
```diff
@@ -182,6 +182,11 @@ M.defaults = {
       { "n", "[x",            actions.prev_conflict,        { desc = "Go to the previous conflict" } },
       { "n", "]x",            actions.next_conflict,        { desc = "Go to the next conflict" } },
       { "n", "g?",            actions.help("file_panel"),   { desc = "Open the help panel" } },
+      { "n", "<leader>co",    actions.conflict_choose_file("ours"),    { desc = "Choose the OURS version of a conflict" } },
+      { "n", "<leader>ct",    actions.conflict_choose_file("theirs"),  { desc = "Choose the THEIRS version of a conflict" } },
+      { "n", "<leader>cb",    actions.conflict_choose_file("base"),    { desc = "Choose the BASE version of a conflict" } },
+      { "n", "<leader>ca",    actions.conflict_choose_file("all"),     { desc = "Choose all the versions of a conflict" } },
+      { "n", "dx",            actions.conflict_choose_file("none"),    { desc = "Delete the conflict region" } },
```
sindrets, line 189: These should not be file panel mappings, because the action doesn't operate on the entry under the cursor. It operates on the currently open file in the view. So these mappings should be moved to the `view` context.

I also think the actual mappings should be:

```
<leader>cO
<leader>cT
<leader>cB
<leader>cA
dX
```

lua/diffview/vcs/utils.lua
```diff
@@ -379,6 +379,5 @@ function M.parse_conflicts(lines, winid)
   return ret, cur_conflict, cur_idx or 0
 end
 
-
```
sindrets, line 382: Unrelated change:  remove.

sindrets (changes requested):
Thanks for the PR! I left some comments. Here's what remains to be done in order to move this forward:

- Your branch is a bit behind: please rebase over current `main`.
- Address the review comments.
- Remove debug print statements.
- Amend docs: document the new action `conflict_choose_all()`.

lua/diffview/actions.lua
```diff
@@ -336,19 +336,83 @@ local function diff_copy_target(kind)
...
+            api.nvim_buf_set_lines(curfile.bufnr, cur.first - 1, cur.last, false, content or {})
+
+            utils.set_cursor(main.id, unpack({
+              (content and #content or 0) + cur.first - 1,
+              content and content[1] and #content[#content] or 0
+            }))
+
+            local lines_count = 0
+            for _ in pairs(content or {}) do lines_count = lines_count + 1 end
+            drift_lines = lines_count - (cur.last - cur.first + 1)
```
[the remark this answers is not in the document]
616b2f: I just thought, do I need to move the cursor at all here? Can you think of any edge case where it could lead to an issue if I just let the cursor where it is?

lua/diffview/config.lua
616b2f: I changed to the mappings you provided, they make more sense.

Having the bindings in the `view` means I can't trigger it from the `file_panel`, which I also would like to be able to do. For me it makes sense to have them in both, same as for `]x` which is also in both.

I added them to both now and can remove if you disagree.

sindrets: Wanting to be able to use the action from the file panel is fine, but you haven't implemented the action as a file panel action. Notably it does not behave like other file panel actions that operate on file entries. Specifically: it doesn't operate on the entry under the cursor in the file panel, but rather *always* on the entry currently open in the view.

As a user I would experience it as highly unexpected behavior if I put the cursor on file `foo` in the file panel, invoked an action, and then had the action operate on `bar` instead, simply because that was the file open in the view.

Our rules for file inference are also clearly defined under `:h diffview-file-inference`.

---

If you want to make it a file panel action it needs to change in a few ways:

- If the panel is *not* focused: the action does exactly what it currently does.
- If the panel *is* focused:
    - Get the current item under the cursor in the panel
    - If it's the currently open file: the action does exactly what it currently does.
    - If it's *not* the currently open file:
        - The file entry needs to be opened
        - We then await the event that signifies that the file is ready (opening file entries is an async operation because we might need to create the files from Git).
        - Then the action proceeds like normal.

To save you the hassle of figuring out the correct API's to call here, I'll give you the file opening logic:

```lua
local view = lib.get_current_view()
if not (view and view:instanceof(DiffView.__get())) then return end
---@cast view DiffView

if view.panel:is_focused() then
  local item = view:infer_cur_file(false) ---@cast item -DirData
  if not item then return end

  if item.active then
    -- The entry is already open and the action can proceed like normal
  else
    -- The entry is not open

    view.emitter:once("file_open_post", function(e, ...)
      -- When this callback is invoked, the entry is ready and the action can
      -- proceed like normal
    end)

    -- Open the entry
    view:set_file(item)
  end
end
```

lua/diffview/actions.lua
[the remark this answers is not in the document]
sindrets: There's never a technical reason to move the cursor here. We move the cursor for UX reasons. It's to communicate to the user that "something changed in this buffer, in this location". With this action we potentially change the buffer in multiple locations, but I think it makes sense to place the cursor in place of the last conflict.

lua/diffview/config.lua
616b2f: You are right, when I think about it I would also expect it to operate on the file under the cursor then on the active view, so I will read the help page you provided and change the implementation to fit the actual use case.

Many thanks for taking time to write me the file opening logic part, really appreciate it!

lua/diffview/actions.lua
```diff
@@ -338,6 +338,91 @@ local function diff_copy_target(kind)
...
+
+        content = content or {}
+        api.nvim_buf_set_lines(curfile.bufnr, first - 1, last, false, content)
+        offset = offset + (#content - (last - first) - 1)
+      end
+
+      utils.set_cursor(main.id, unpack({
+        (content and #content or 0) + first - 1,
+        content and content[1] and #content[#content] or 0
+      }))
```
sindrets, line 384: Use the `sync_scroll()` method here, as scroll-bound windows are not synchronized if one of their cursors move while its window is unfocused.

```suggestion
      }))

      view.cur_layout:sync_scroll()
```

lua/diffview/actions.lua
```diff
@@ -338,6 +338,91 @@ local function diff_copy_target(kind)
...
+          view.emitter:once("file_open_post", function()
+            -- When this callback is invoked, the entry is ready and the action can
+            -- proceed like normal
+            resolve_all_conflicts(view, target)
+          end)
+
+          -- Open the entry
+          view:set_file(item)
+        end
+      end
```
sindrets, line 415:
```suggestion
      else
        resolve_all_conflicts(view, target)
      end
```

lua/diffview/actions.lua
```diff
@@ -338,6 +338,91 @@ local function diff_copy_target(kind)
...
+    end
+
+    if view and view:instanceof(StandardView.__get()) then
+      ---@cast view StandardView
+
+      resolve_all_conflicts(view, target)
+    end
```
sindrets, line 422: Remove. `StandardView` is a super class of `DiffView`.

sindrets (changes requested):
Great! Just some small things and then this should be good to go!

616b2f:
done. Thank you so much, for taking the time to help me get this thing done. I learned a lot.

sindrets (approved):
LGTM, thank you for your contribution!

PR #323, opened 2023-03-18
feat(file-history): Hide unrelated changes when tracing line evolution

sindrets:
When tracing line evolution (the `-L` flag option) we look at the diff patches Git outputs in order to determine what the relevant content is. We then hide the irrelevant content by setting `foldmethod=manual` in all the diff buffers and creating custom folds derived from the hunks in the git diff.

This seems to work quite well in most cases, and makes it a lot more efficient to review changes with the `-L` flag, as we no longer need to guess where in the file the change occurred. After all, Git knows this far better than us.

Bonus change: the `scroll_view()` action now uses the window that currently has the most visible content, to scroll. This makes the scrolling far less jumpy in cases where you have a large deleted hunk. As the cursor cannot be placed inside a region of filler lines, the scroll action would previously just shoot past such regions, something that could be very disorienting. Now it will instead be able to tell that the other window has more visible content, and scroll that instead.

Also fixed the bug where `scroll_view()` with a negative number would move the cursor up one line every time it was invoked.

### DIFF
--- a/lua/diffview/actions.lua
+++ b/lua/diffview/actions.lua
@@ -29,6 +29,8 @@ local M = setmetatable({}, {
   end
 })
 
+M.compat = {}
+
 ---@return FileEntry?
 ---@return integer[]? cursor
 local function prepare_goto_file()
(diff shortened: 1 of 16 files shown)


weilbith:
Thank you very much. That is amazing 😊

PR #326, opened 2023-03-24
use `nvim_get_hl` if available

3699394:
use latest api if available https://github.com/neovim/neovim/pull/22693

### DIFF
--- a/lua/diffview/hl.lua
+++ b/lua/diffview/hl.lua
@@ -121,12 +121,12 @@ function M.get_hl(name, no_trans)
   local hl
 
   if no_trans then
-    hl = api.nvim__get_hl_defs(0)[name]
+    hl = api.nvim__get_hl_defs and api.nvim__get_hl_defs(0)[name] or api.nvim_get_hl(0, {name=name})
   else
     local id = api.nvim_get_hl_id_by_name(name)
 
     if id then
-      hl = api.nvim_get_hl_by_id(id, true)
+      hl = api.nvim_get_hl_by_id and api.nvim_get_hl_by_id(id, true) or api.nvim_get_hl(0, {id=id})
     end
   end
 


cleong14 (commented):
Tested these changes locally and can confirm they fixed #327 for me.

mikesmithgh:
> Tested these changes locally and can confirm they fixed #327 for me.

Awesome, thanks!

lua/diffview/hl.lua
```diff
@@ -121,12 +121,12 @@ function M.get_hl(name, no_trans)
   local hl
 
   if no_trans then
-    hl = api.nvim__get_hl_defs(0)[name]
+    hl = api.nvim__get_hl_defs and api.nvim__get_hl_defs(0)[name] or api.nvim_get_hl(0, {})[name]
```
mehalter, line 124:
```suggestion
    hl = api.nvim__get_hl_defs and api.nvim__get_hl_defs(0)[name] or api.nvim_get_hl(0, {name=name})
```

This might be a cleaner implementation

sindrets:
Thanks for the PR! I don't have time to properly review this for about a week. But after a quick look it seems like the new API returns slightly different data than the old `nvim__get_hl_defs()` (some of the keys have been renamed). So although your suggested change might avoid the error, it might not be the correct way to accommodate this new API. I will have to look closer into the matter when I have time.

mehalter:
@sindrets this is correct, to help with some insight. The old approach returns properties `foreground` and `background`, the new API returns these as `fg` and `bg`

lua/diffview/hl.lua
3699394: Thanks for the recommendation, overlooked when I committed. Updated the pr

TisnKu:
Thanks for the PR。

gegoune:
@mosheavni Be patient. https://github.com/sindrets/diffview.nvim/pull/326#issuecomment-1484093994
Use your package manager to pin to latest working version in meantime.

jamestrew:
You can subscribe to this PR to get notified of when it gets merged as well.

Breaking bugs like this is par for the course when riding neovim nightly so let's exercise some patience. You can always switch back to stable if you want to.

mikesmithgh:
> You can subscribe to this PR to get notified of when it gets merged as well.
> 
> Breaking bugs like this is par for the course when riding neovim nightly so let's exercise some patience. You can always switch back to stable if you want to.

100% agree. If you are on nightly, you should expect occasional breaking changes.

c3n21:
> Guys, what's the hold up? the plugin is completely broken now. Can this be merged already? Thanks.

While the maintainer is merging this PR if you are using `bob` you can pin this version `ea0b66d` which is the commit before the one that introduces this change

3699394:
> Guys, what's the hold up? the plugin is completely broken now. Can this be merged already? Thanks.

you can patch `hl.lua` with my pr, that's what I have been doing

Issue #111, opened 2022-01-24
Go to next difference


tristone13th:
Is it possible to go to the next different place between the two files?

gegoune:
Try `[c` and `]c` It works for me although `:map [c` shows 'No mapping found'.

sindrets:
Yes, exactly what @gegoune said. That's a default binding in vim. It's documented under `:h jumpto-diffs`.

tristone13th:
It worked! Many thanks to you all.

louwers:
Is it possible to automatically move to the next file?

sindrets:
@louwers with `]c`? No. You could do something like this I suppose:

```lua
require("diffview").setup({
  keymaps = {
    view = {
      {
        "n", "]c",
        function ()
          local cursor = vim.api.nvim_win_get_cursor(0)
          vim.cmd("norm! ]c")
          if cursor[1] == vim.api.nvim_win_get_cursor(0)[1] then
            -- The cursor didn't move: go to the next entry
            actions.select_next_entry()
          end
        end,
        { desc = "Next hunk / entry" },
      },
    },
  },
})
```

Keep in mind that if you're using gitsigns with its suggested mappings, it will override this in its `on_attach` callback.

louwers:
Cool 😎

Issue #277, opened 2022-12-28
Mention `:h diff-mode` in the README


sathishmanohar:
`]c` jumps to the next hunk and `[c` jumps to the previous hunt but in the `g?` help section it isn't listed.

sindrets:
Yes, because they are builtin mappings: `:h jumpto-diffs`. The plugin assumes you're familiar with everything from `:h diff-mode`.

sathishmanohar:
Got it. This can be mentioned in the readme somewhere

PR #334, opened 2023-03-30
docs: Explain how to jump between diffs

MatejKastak:
Hello, very good plugin :+1:

I am starting to use it for more complex Pull Request reviews. But I was not able to easily figure out how to jump to next difference. I tried to check README for `next*` and saw `next_conflict()` and it was doing nothing for me. Only after actually reading the code I finally accepted that it was doing exactly what was the name. Nevertheless, it was the only `next_*` reference in the plugin so I did not dismiss it sooner. 

I read the whole README.md, some of the docs, keyboard mappings configuration, etc. but it took me some time to stumble upon an old issue https://github.com/sindrets/diffview.nvim/issues/111 which was exactly what I was looking for. So I decided to extend the FAQ (feel free to edit it in any way). Judging by the number of reactions on that issues other people are looking for it also.

Hopefully, it will be useful for others :slightly_smiling_face: and again thanks for a cool plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,3 +467,7 @@ git object database, and a command is echoed that shows how to undo the change.
     - (Lua): `vim.opt.fillchars:append { diff = "╱" }`
   - Note: whether or not the diagonal lines will line up nicely will depend on
     your terminal emulator. The terminal used in the screenshots is Kitty.
+- **Q: How do I jump to next diff (not next conflict)?**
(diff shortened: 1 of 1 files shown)


sindrets:
Eventually I'll remove this and replace it with a section about `:h diff-mode` (rel. #277), but for now: LGTM, thanks!

PR #347, opened 2023-04-21
fix: use correct diff layout

zegervdv:
When running `DiffviewOpen` on a merge conflict I get the following error:
``` 
The coroutine failed with this message: ...nvim/lua/diffview/scene/views/standard/s
tandard_view.lua:143: attempt to index a nil value                                                                                                                                            
stack traceback:                                                                                                                                                                              
        [C]: in function 'error'                                                                                                                                                              
        ...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:18: in function 'callback_or_next'                                                                                        
        ...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:45: in function <...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:44> 
```

I've bisected the repo and found 5fc2b59e9668c06588e9f0cacb79d459067ae273 to have introduced this.

Seems like the `layout_key` can be `nil` when opening the view.

### DIFF
--- a/lua/diffview/scene/views/standard/standard_view.lua
+++ b/lua/diffview/scene/views/standard/standard_view.lua
@@ -135,7 +135,6 @@ function StandardView:use_entry(entry)
     layout_key = "diff4"
   end
 
-
   for _, sym in ipairs({ "a", "b", "c", "d" }) do
     if entry.layout[sym] then
       entry.layout[sym].file.winopts = vim.tbl_extend(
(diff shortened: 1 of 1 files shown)


sindrets:
Seems like the actual root of the problem here is that I'm importing the wrong thing here:

https://github.com/sindrets/diffview.nvim/blob/3733cdad0ec88e6f5e6a07d6e9cf488c87a82d38/lua/diffview/scene/views/standard/standard_view.lua#L6

`"...layouts.diff_3"` should be `"...layouts.diff_4"`

So we've never actually been checking whether the layout was a `Diff4` here.

zegervdv:
updated

sindrets (approved):
Great, thank you so much!

PR #370, opened 2023-06-05
feat(file-history): Added `--before` and `--after`

emmanueltouzery:
support for git log flags.

### DIFF
--- a/lua/diffview/vcs/adapters/git/init.lua
+++ b/lua/diffview/vcs/adapters/git/init.lua
@@ -766,8 +766,8 @@ function GitAdapter:file_history_options(range, paths, argo)
     { "base" },
     { "G" },
     { "S" },
-    { "after" },
-    { "before" },
+    { "after", "since" },
+    { "before", "until" },
   }
 
   local log_options = { rev_range = range_arg } --[[@as GitLogOptions ]]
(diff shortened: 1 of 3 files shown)


sindrets (changes requested):
Thanks for the PR! This looks pretty good. Please just amend [the type definition for `GitLogOptions`](https://github.com/emmanueltouzery/diffview.nvim/blob/df507ddea4f064f1a750f429230ebe373ba196fe/lua/diffview/config.lua#L254) and the docs for `:h :DiffviewFileHistory`.

sindrets:
In the docs it could also be worth mentioning that the flags accept dates in natural language as well (like i.e. `"2 weeks 6 hours 12 minutes ago"`).

lua/diffview/vcs/adapters/git/init.lua
```diff
@@ -1787,6 +1791,12 @@ GitAdapter.flags = {
     FlagOption("=S", "-S", "Search occurrences", {
       prompt_label = "(Extended regular expression)"
     }),
+    FlagOption("=A", "--after==", "List only commits after a certain date", {
```
sindrets, line 1794: Should only be one equal sign: `--after=`

emmanueltouzery:
applied the requested changes :+1:

sindrets (approved):
LGTM, thank you! I made a small change adding Git's synonyms for these flags `--since` and `--until`.

emmanueltouzery:
:+1: on the since & until.

PR #374, opened 2023-06-08
fix: Workaround for `<afile>` expanding incorrectly

### LINKED ISSUE #369: `TabClosed` callbacks break if `expand("<afile>")` is called prior
Is this an expected behavior or a bug? Because after exploring a little deeper, I found that closing multiple panels is done via `TabClosed` event callbacks, defined here

https://github.com/sindrets/diffview.nvim/blob/86bf6182e2ea767c21711de8a3c396d9b635d970/lua/diffview/init.lua#L44-L50

However, the 'TabClosed' event is specified to be non-recursive (event cannot trigger itself) (I didn't find this in the `:h tabclosed`, but I found this logic in the neovim source code [static void do_autocmd_winclosed(win_T *win)](https://github.com/neovim/neovim/blob/fdc8e966a9183c08f2afec0817d03b7417a883b3/src/nvim/window.c#L3018C45-L3030)), so the end result is that the TabClosed event will not be triggered again when the close operation is performed a second time and the remaining panels will not be closed further down.

Now, I got around this by forcing the event to fire again by calling vim.api.nvim_exec_autocmds in the TabClosed callback, but I'm not sure if that's the right solution, or if the result (only two panels closed) is what you expect.

A small addition, if this is a bug, can you try to let me fix it? I really like this plugin and would like to give some modest contribution.

NGPONG:
Resolves #369.

I have left some question here:

1. Added a comment not related to this pr, sorry if this makes you uncomfortable I will restore it.

2. I'm not sure if the fetch_tabnr's *name* is correct.

3. I'm not sure if fetch_tabnr is in the right place for this, I tried to throw it into *utils* but its logic is not that common.

### DIFF
--- a/lua/diffview/init.lua
+++ b/lua/diffview/init.lua
@@ -23,6 +23,22 @@ end
 
 function M.init()
   local au = api.nvim_create_autocmd
+
+  -- Fix the strange behavior that "<afile>" expands non-files
+  -- as file name in some cases.
+  --
+  -- About more information please checkout:
+  --  * sindrets/diffview.nvim#369
+  --  * neovim/neovim#23943
+  local fetch_tabnr = function (state)
+    if vim.fn.has("nvim-0.9.2") ~= 1 then
+      return tonumber(state.match)
+    else
+      return tonumber(state.file)
+    end
+  end
+
+  -- Set up highlighting
   hl.setup()
 
   -- Set up autocommands
@@ -45,7 +61,7 @@ function M.init()
     group = M.augroup,
     pattern = "*",
     callback = function(state)
-      M.close(tonumber(state.file))
+      M.close(fetch_tabnr(state))
     end,
   })
   au("BufWritePost", {
@@ -59,7 +75,7 @@ function M.init()
     group = M.augroup,
     pattern = "*",
     callback = function(state)
-      M.emit("win_closed", tonumber(state.file))
+      M.emit("win_closed", fetch_tabnr(state))
     end,
   })
   au("ColorScheme", {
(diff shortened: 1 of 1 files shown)


lua/diffview/init.lua
```diff
@@ -23,6 +23,22 @@ end
 
 function M.init()
   local au = api.nvim_create_autocmd
+
+  -- Fix the strange behavior that "<afile>" expands non-files
+  -- as file name in some cases.
+  --
+  -- About more information please checkout:
```
sindrets, line 30:
```suggestion
  -- Ref:
```

lua/diffview/init.lua
```diff
@@ -23,6 +23,22 @@ end
...
+  --  * sindrets/diffview.nvim#369
+  --  * neovim/neovim#23943
+  local fetch_tabnr = function (state)
```
sindrets, line 33: Let's just call it `get_tabnr`.

Nit pick: I prefer using Lua's syntactic sugar for defining functions. It's a little denser and it makes a clearer distinction between functions and other types.

```suggestion
  local function get_tabnr(state)
```

sindrets (changes requested):

lua/diffview/init.lua
NGPONG: fixed

lua/diffview/init.lua
NGPONG: Great shout! This does get better and has been fixed.

sindrets (approved):
LGTM, thank you!

PR #376, opened 2023-06-09
Update USAGE.md

briandipalma:
Support default branches that aren't called `main`.


sindrets (changes requested):
Please update the other references to `origin/main` in the text. And also make the same changes in the next section.

briandipalma:
The italics changes were made by Prettier I'm guessing? I've no idea why underscores are preferred over asterisks, it seemed innocuous so I left the changes.

sindrets (approved):
LGTM, thank you! I added a little explanation for how to get Git to set `origin/HEAD`, as it isn't always set by default (i.e. bare repos).

PR #377, opened 2023-06-09
Correct `enhanced_diff_hl` documentation

vinnydiehl:
I needed to do this for my colorscheme, figured I'd help the next guy out.

### DIFF
--- a/doc/diffview.txt
+++ b/doc/diffview.txt
@@ -361,10 +361,11 @@ enhanced_diff_hl                                *diffview-config-enhanced_diff_h
 
         Enable/disable enhanced diff highlighting. In `diff2*` layouts this
         will highlight the delete fill-chars more subtly, and show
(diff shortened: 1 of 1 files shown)


sindrets:
The documentation describes the setting's functional effect. How the plugin achieves that function is an implementation detail, and does not belong in the user documentation. This change makes the docs less clear to the user.

```vimhelp
In `diff2*` layouts this will highlight the delete fill-chars more subtly, and
show |hl-DiffAdd| in the old state buffer, as |hl-DiffviewDiffAddAsDelete|.
```

This line means nothing to a user. What is `DiffviewDiffAddAsDelete`? In the current docs the user can follow the tag link to the documentation for `|hl-DiffDelete|` in order to read about what that diff highlight applies to.

`DiffviewDiffAddAsDelete` inherits its values from `DiffDelete`, so as a color scheme creator you should only be changing `DiffDelete` regardless.

vinnydiehl:
Ah. My mistake, cheers.

PR #424, opened 2023-10-06
feat: remove bottom extra blank line in file panel

0x00-ketsu:
In Diffview panel, Press command `G`, it'is jump to bottom line, which is a empty line.

image

### DIFF
--- a/lua/diffview/renderer.lua
+++ b/lua/diffview/renderer.lua
@@ -508,6 +508,10 @@ function M.render(bufid, data)
     hl_data = { data.hl }
   end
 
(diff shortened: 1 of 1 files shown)


sindrets (changes requested):
If you're gonna change this, don't change logic in the renderer. The renderer should always render the data it's given, as is. You want to change the UI logic, around here: https://github.com/sindrets/diffview.nvim/blob/0437ef8bfdd67156d87140d692840a3c2824fa20/lua/diffview/scene/views/diff/render.lua#L134

PR #425, opened 2023-10-07
docs: Add commit example using `vim.ui.input()`

LintaoAmons:
A more intuitive way of commit. Just one keystroke after stash right in diffview

show

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -208,6 +208,9 @@ keymaps = {
 
 This will popup a simple input box for your to enter the commit msg and commit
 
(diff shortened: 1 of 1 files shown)


sindrets:
Thanks! I made some adjustments so that it doesn't need to rely on Plenary. Nvim has plenty of builtin ways of running and controlling jobs. Unfortunately, many of the interfaces we have for managing jobs are either awkward to use, or too low level to use easily. But the new `vim.system()` is quite nice.

LintaoAmons:
Haha, happy to see this is accepted~

PR #453, opened 2024-01-03
feat: Add the DiffviewToggle command

benwilliams140:
Use `:DiffviewToggle` to open/close the Diffview. Accepts the same arguments as the `:DiffviewOpen` command.


sindrets:
Thanks for the PR, but this has been discussed before, and it's not something I want because I think it's conceptually flawed. Please read discussions in parent issues.

PR #463, opened 2024-02-18
fix(git): fix data access if nil

lucobellic:
I had an error while using `DiffViewFileHistory` with nvim nightly v0.10.0-dev-2361+ga376d979b

The command now works as expected.


sindrets:
Thanks for the PR!

I'm missing a proper bug report for the issue described in this PR. I'm hesitant to consider including this change as it's not really a fix. It's adding special handling for unexpected behavior. Something that is generally a bad idea. This is merely pushing the root of the problem under a rug.

I'm gonna close this PR at least until I see a bug report.

PR #465, opened 2024-02-20
docs: Add lazy.nvim to installation docs

woodevin:


josealvaradoo (commented):
Great 👌🏻

Balake:
@woodevin I had to do `lazy = false`, not sure if that's true for all installations.

amiv1:
> @woodevin I had to do `lazy = false`, not sure if that's true for all installations.

It's better to specify either event or a command to make it load on demand, instead of during nvim startup.
```lua
{
  "sindrets/diffview.nvim",
  cmd = { "DiffviewOpen" },
}
```

Docs: https://lazy.folke.io/spec#spec-lazy-loading

PR #472, opened 2024-03-10
fix: cancel update if view becomes out of focus

### LINKED ISSUE #462: [Bug] CTRL-W_T does not move focus to new tab
### Description

In the first place, thank you for your amazing plugin! Either on its own or utilized by my other favourite plugin [gitlab.nvim](https://github.com/harrisoncramer/gitlab.nvim), `diffview.nvim` is of enormous help in reviewing code history.

Quite frequently, while inspecting a diff with `diffview`, I open a file in a new split, e.g,. a help page, and then I decide to move that window to another tab page. This can be achieved with the `<c-w>T` normal mode command (see `:h CTRL-W_T`). This command first moves the window to a new tab page and then moves focus to that tab. However, when in a `diffview` tab, focus stays in this tab while the screen flickers as the view is switched several times between the newly created tab page and the `diffview` tab.

### Expected behavior

CTRL-W_T moves current window to a new tab page and switches focus to that tab

### Actual behavior

CTRL-W_T moves current window to a new tab page and switches focus to that tab and back to the `diffview` tab several times.

### Steps to reproduce

1. `nvim --clean -u mini.lua`
2. `:DiffviewOpen`
3. `:help CTRL-W_T`
4. `<c-w>T`

### Health check

<details>
<summary>Output of <code>:checkhealth diffview</code></summary>

```
diffview: require("diffview.health").check()

Checking plugin dependencies ~
- OK nvim-web-devicons installed.

Checking VCS tools ~
- The plugin requires at least one of the supported VCS tools to be valid.
- OK Git found.
- OK Git is up-to-date. (2.43.0)
- WARNING Configured `hg_cmd` is not executable: 'hg'
```

</details>

### Log info

_No response_

### Neovim version

```markdown
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1702233742
```

### Operating system and version

Linux 6.1.71-1-MANJARO x86_64 GNU/Linux

### Minimal config

```Lua
-- #######################################
-- ### USAGE: nvim --clean -u mini.lua ###
-- #######################################

local root = vim.fn.stdpath("run") .. "/nvim/diffview.nvim"
vim.fn.mkdir(root, "p")

-- set stdpaths to use root
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  {
    "sindrets/diffview.nvim",
    dependencies = { "nvim-tree/nvim-web-devicons" },
    config = function()
      require("diffview").setup({
        -- ##############################################################################
        -- ### ADD DIFFVIEW.NVIM CONFIG THAT IS _NECESSARY_ FOR REPRODUCING THE ISSUE ###
        -- ##############################################################################
      })
    end,
  },
  -- ##################################################################
  -- ### ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE ###
  -- ##################################################################
}
require("lazy").setup(plugins, { root = root .. "/plugins" })
require("lazy").sync({ show = false, wait = true })

vim.opt.termguicolors = true
vim.cmd("colorscheme " .. (vim.fn.has("nvim-0.8") == 1 and "habamax" or "slate"))

-- ############################################################################
-- ### ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE ###
-- ############################################################################
```

8veStory:
Fixes #462.

I work in a large monorepo and noticed that when :DiffviewOpen was used, and you switched to another tab, you would switch back to the Diffview window.

Reason for this being that inside the async function `DiffView:update_file()`, it is possible for the line `local err, new_files = await(self:get_updated_files())` to take a long amount of time in large git repos. This meant it's possible for a user to switch to another tab prior to the async function completing, in which at the end of the function the user's window is switched to the DiffView.

I'm not entirely familiar with Lua but I made a very rough PR that changes the order of when re-rendering the panels work to abate this issue.

### DIFF
--- a/lua/diffview/scene/views/diff/diff_view.lua
+++ b/lua/diffview/scene/views/diff/diff_view.lua
@@ -327,6 +327,12 @@ DiffView.update_files = debounce.debounce_trailing(
   async.wrap(function(self, callback)
     await(async.scheduler())
 
+    -- Never update unless the view is in focus
+    if self.tabpage ~= api.nvim_get_current_tabpage() then
+      callback({ "The update was cancelled." })
+      return
+    end
+
     ---@type PerfTimer
     local perf = PerfTimer("[DiffView] Status Update")
     self:ensure_layout()
@@ -344,16 +350,21 @@ DiffView.update_files = debounce.debounce_trailing(
     end
 
     local index_stat = pl:stat(pl:join(self.adapter.ctx.dir, "index"))
-    local last_winid = api.nvim_get_current_win()
 
     ---@type string[]?, FileDict
     local err, new_files = await(self:get_updated_files())
+    await(async.scheduler())
 
     if err then
       utils.err("Failed to update files in a diff view!", true)
       logger:error("[DiffView] Failed to update files!")
-      await(async.scheduler())
       callback(err)
+      return
+    end
+
+    -- Stop the update if the view is no longer in focus.
+    if self.tabpage ~= api.nvim_get_current_tabpage() then
+      callback({ "The update was cancelled." })
       return
     end
 
@@ -364,8 +375,6 @@ DiffView.update_files = debounce.debounce_trailing(
       { cur_files = self.files.working, new_files = new_files.working },
       { cur_files = self.files.staged, new_files = new_files.staged },
     }
-
-    await(async.scheduler())
 
     for _, v in ipairs(files) do
       -- We diff the old file list against the new file list in order to find
@@ -472,10 +481,6 @@ DiffView.update_files = debounce.debounce_trailing(
       end
     end
     self:set_file(self.panel.cur_file or self.panel:next_file(), false, not self.initialized)
-
-    if api.nvim_win_is_valid(last_winid) then
-      api.nvim_set_current_win(last_winid)
-    end
 
     self.update_needed = false
     perf:time()
--- a/lua/diffview/scene/views/diff/diff_view.lua
+++ b/lua/diffview/scene/views/diff/diff_view.lua
@@ -344,6 +344,7 @@ DiffView.update_files = debounce.debounce_trailing(
     end
 
     local index_stat = pl:stat(pl:join(self.adapter.ctx.dir, "index"))
+    local last_winid = api.nvim_get_current_win()
 
     ---@type string[]?, FileDict
     local err, new_files = await(self:get_updated_files())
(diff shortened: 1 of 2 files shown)


sindrets (commented):
Thanks for the PR!

The plugin probably needs a bigger refactor to better deal with synchronization issues. But in this case we should be able to improve the behavior considerably with a smaller change I think.

The problem isn't with rendering. Panels can render without changing focus. I think we should just cancel the file update if the view is no longer in focus when `get_updated_files()` settles. We can determine that by looking at the current tab page before and after awaiting this method. How about something like this:

```diff
diff --git a/lua/diffview/scene/views/diff/diff_view.lua b/lua/diffview/scene/views/diff/diff_view.lua
index 283fed9..84d9941 100644
--- a/lua/diffview/scene/views/diff/diff_view.lua
+++ b/lua/diffview/scene/views/diff/diff_view.lua
@@ -327,6 +327,12 @@ DiffView.update_files = debounce.debounce_trailing(
   async.wrap(function(self, callback)
     await(async.scheduler())
 
+    -- Never update unless the view is in focus
+    if self.tabpage ~= api.nvim_get_current_tabpage() then
+      callback({ "The update was cancelled." })
+      return
+    end
+
     ---@type PerfTimer
     local perf = PerfTimer("[DiffView] Status Update")
     self:ensure_layout()
@@ -344,19 +350,24 @@ DiffView.update_files = debounce.debounce_trailing(
     end
 
     local index_stat = pl:stat(pl:join(self.adapter.ctx.dir, "index"))
-    local last_winid = api.nvim_get_current_win()
 
     ---@type string[]?, FileDict
     local err, new_files = await(self:get_updated_files())
+    await(async.scheduler())
 
     if err then
       utils.err("Failed to update files in a diff view!", true)
       logger:error("[DiffView] Failed to update files!")
-      await(async.scheduler())
       callback(err)
       return
     end
 
+    -- Stop the update if the view is no longer in focus.
+    if self.tabpage ~= api.nvim_get_current_tabpage() then
+      callback({ "The update was cancelled." })
+      return
+    end
+
     perf:lap("received new file list")
 
     local files = {
@@ -365,8 +376,6 @@ DiffView.update_files = debounce.debounce_trailing(
       { cur_files = self.files.staged, new_files = new_files.staged },
     }
 
-    await(async.scheduler())
-
     for _, v in ipairs(files) do
       -- We diff the old file list against the new file list in order to find
       -- the most efficient way to morph the current list into the new. This
@@ -473,10 +482,6 @@ DiffView.update_files = debounce.debounce_trailing(
     end
     self:set_file(self.panel.cur_file or self.panel:next_file(), false, not self.initialized)
 
-    if api.nvim_win_is_valid(last_winid) then
-      api.nvim_set_current_win(last_winid)
-    end
-
     self.update_needed = false
     perf:time()
     logger:lvl(5):debug(perf)
```

sindrets:
@8veStory sorry for hijacking your PR. I think this should solve the issue you described, but let me know if you still have any problems, and we can iterate on this.

PR #484, opened 2024-04-13
fix(git): add `--no-exclude-standard` when detecting binary

Summon528:
Otherwise, files that was included in gitignore (but added via `git add --force`) will be treated as binary and not showing diff.


sindrets (changes requested):
Thanks for the PR! This change makes sense, just fix that one bug and I'll include this.

Summon528:
Updated, thanks

sindrets (approved):
LGTM, thanks!

Issue #496, opened 2024-05-20
[Feature request] Have an option to use gerrit style file sort.


eternity74:
Hi, Thanks for providing great plugin for daily use.

I sometimes compare local commit with gerrit changes by using diffview.nvim plugin.
For this use-case, there is mismatch on file sort compare to gerrit.
Gerrit seems to give higher priority to cppheader files so header file comes first and source file laster.

I want to have the same behavior with option like gerrit_style_filesort = true.

PR #497, opened 2024-05-20
feat: Support gerrit style file sort. (#496)

eternity74:
This patch allows to set gerrit style file sort (give higher priority to c/cpp header files) by setting gerrit_style_filesort = true Default value is false.


dlyongemallo:
@eternity74 This can be implemented using the `file_panel.sort_file` config option in [my actively maintained fork](https://github.com/dlyongemallo/diffview.nvim) of this repo.

PR #499, opened 2024-05-21
fix: `vim.diagnostic.disable` is deprecated

yorickpeterse:
Instead one should use vim.diagnostic.enable in the following form:

    vim.diagnostic.enable(true | false, { bufnr = BUFFER })

### DIFF
--- a/lua/diffview/vcs/file.lua
+++ b/lua/diffview/vcs/file.lua
@@ -15,6 +15,8 @@ local pl = lazy.access(utils, "path") ---@type PathLib
 local api = vim.api
 local M = {}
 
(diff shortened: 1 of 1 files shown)


sindrets (changes requested):
Thanks for the PR!

This change isn't backwards compatible. We should use `:h has()` to detect the feature level of the client and then use the appropriate API depending on the result.

yorickpeterse:
@sindrets Thanks for the reminder, I've adjusted the changes to check for the NeoVim version based on how this is done in some other places :smiley:

sindrets (approved):
LGTM, thanks!

PR #507, opened 2024-05-29
feat: use `vim.notify` for user notifications

### LINKED ISSUE #473: [feat][would write PR] use vim.notify for user notifications
I would be interested to write a PR for that, but I'm afraid that you'd reject it, so I would first ask if you'd agree with the approach.

It seems to me that in particular error messages (eg `utils.err()` in the source) are not that emphasized in the GUI. neovim introduced the vim.notify call that can be overridden by users for custom display. For instance the vim.notify plugin offers such an override:
https://github.com/rcarriga/nvim-notify

I would like to enable the user to decide between the current display (using `nvim_echo`) or vim.notify, that would allow the user to customize the display.

Possible options...
1. rewrite utils.info/warn/err to directly call vim.notify, and delete `echo_multiln`
2. add a new config option, for instance `user_messages_with_notify = false`
3. add new hooks for user messages, for instance `user_message_displayed`. Although that's a bit redundant, because vim.notify already allows customization but...

If you think such a feature has its place in diffview, feel free to let me know which approach you'd rather see, and I can prepare a PR to that effect!

Thank you for an amazing plugin!

emmanueltouzery:
fixes #473
vim.notify allows the user to customize the display of messages

I'm not sure about the schedule parameter, whether it's still needed. I left it in just in case.

The testing I performed is that I triggered the:
> "The file is open with unsaved changes! Aborting file restoration."

message, by modifying a file, saving, modifying again, then asking to revert the changes in diffview, that triggered that message. I also tried modifying the location where the message is triggered to ask to trigger it with the schedule parameter being "true" and that worked too.

### DIFF
--- a/lua/diffview/utils.lua
+++ b/lua/diffview/utils.lua
@@ -23,65 +23,44 @@ function M.now()
   return vim.loop.hrtime() / 1000000
 end
 
----Echo string with multiple lines.
 ---@param msg string|string[]
----@param hl? string Highlight group name.
 ---@param schedule? boolean Schedule the echo call.
-function M.echo_multiln(msg, hl, schedule)
+---@param level integer
+function M.notify(msg, level, schedule)
   if schedule then
-    vim.schedule(function() M.echo_multiln(msg, hl, false) end)
+    vim.schedule(function() M.notify(msg, level, false) end)
+    return
+  end
+  if type(msg) == "table" then
+    msg = table.concat(msg, "\n")
+  end
+  if msg == "" then
     return
   end
 
-  if type(msg) ~= "table" then msg = { msg } end
+  if level == vim.log.levels.ERROR then
+    logger:error(msg)
+  end
 
-  local text = table.concat(msg, "\n")
-  api.nvim_echo({ { text, hl } }, true, {})
+  vim.notify(msg, level, { title = "diffview.nvim" })
 end
 
 ---@param msg string|string[]
 ---@param schedule? boolean Schedule the echo call.
 function M.info(msg, schedule)
-  if type(msg) ~= "table" then
-    msg = { msg }
-  end
-  if not msg[1] or (msg[1] == "" and #msg == 1) then
-    return
-  end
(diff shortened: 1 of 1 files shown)


lua/diffview/utils.lua
```diff
@@ -23,65 +23,47 @@ function M.now()
...
+---@param number level
+function M.notify(msg, level, schedule)
   if schedule then
-    vim.schedule(function() M.echo_multiln(msg, hl, false) end)
+    vim.schedule(function() M.notify(msg, level, false) end)
+    return
+  end
+  if not msg then
+    return
+  end
```
sindrets, line 36: Remove this, this function should never be called without a `msg` argument. Unintended usage should not be handled. If anything, do `assert(msg ~= nil)`.

sindrets (changes requested):
Just remove that one early return, and this looks good!

And, yes, let's keep the `schedule` parameter. It's useful for two reasons:

1. When these are called from an async context. Without scheduling the notifications are likely to fail when they try to access the nvim API.
2. In performance critical code it's useful for deferring the handling of the notification until later.

emmanueltouzery:
I force pushed the removal the nil-check now. I had it in because the previous code also had it (well, it had `if not msg[1]`). Thank you for the explanation for schedule!

lua/diffview/utils.lua
```diff
@@ -23,65 +23,44 @@ function M.now()
   return vim.loop.hrtime() / 1000000
 end
 
----Echo string with multiple lines.
 ---@param msg string|string[]
----@param hl? string Highlight group name.
 ---@param schedule? boolean Schedule the echo call.
-function M.echo_multiln(msg, hl, schedule)
+---@param number level
```
sindrets, line 28: Nit:

```suggestion
---@param level integer
```

emmanueltouzery:
my bad about the type signature :grimacing: 
force pushed the fix now.

sindrets (approved):
LGTM, thank you!

PR #510, opened 2024-06-11
feat(actions): added `select_{first|last}_entry`

### LINKED ISSUE #421: [feature] Add actions for selecting first / last entry
Hello, I use diffview in a way that I select the current file when opening, using `--selected-file=`. This works great for me.

However sometimes I want to use diffview to review all the changes. So, I open the diff view, the current file selected, but I'd like to jump to the first file to go through all the files one by one and be sure I didn't miss any of them. I don't know a way to do that... I know of:
```
lua require'diffview.config'.actions.select_next_entry()
```

And the equivalent for the previous entry, but how could I jump to the first entry with a single keyboard shortcut? Thank you!

yungthai:
Resolves #421 

There are some extraneous formatting committed (into single lines) due to the following setting within `stylua.toml` below. StyLua was used.
```
collapse_simple_statement = "Always"
```

After the above formatter option was introduced, I had expected  short `if` blocks to be formatted onto single lines.

Would you please confirm whether StyLua is used as the Lua formatter? I'll tidy up those unneccessary formatting once you confirm the formatter tool & its options.


sindrets:
This looks vastly overcomplicated for what should essentially just be `view:set_file(files[1])` and `view:set_file(files[#files])`.

Something like this (untested):

```diff
diff --git a/lua/diffview/actions.lua b/lua/diffview/actions.lua
index 07e3c55..91c8178 100644
--- a/lua/diffview/actions.lua
+++ b/lua/diffview/actions.lua
@@ -635,6 +635,8 @@ local action_names = {
   "select_entry",
   "select_next_entry",
   "select_prev_entry",
+  "select_first_entry",
+  "select_last_entry",
   "stage_all",
   "toggle_files",
   "toggle_flatten_dirs",
diff --git a/lua/diffview/scene/views/diff/listeners.lua b/lua/diffview/scene/views/diff/listeners.lua
index e0876aa..1fb8b83 100644
--- a/lua/diffview/scene/views/diff/listeners.lua
+++ b/lua/diffview/scene/views/diff/listeners.lua
@@ -71,6 +71,14 @@ return function(view)
     select_prev_entry = function()
       view:prev_file(true)
     end,
+    select_first_entry = function()
+      local files = view.panel:ordered_file_list()
+      if files and #files > 0 then view:set_file(files[1]) end
+    end,
+    select_last_entry = function()
+      local files = view.panel:ordered_file_list()
+      if files and #files > 0 then view:set_file(files[#files]) end
+    end,
     next_entry = function()
       view.panel:highlight_next_file()
     end,
diff --git a/lua/diffview/scene/views/file_history/listeners.lua b/lua/diffview/scene/views/file_history/listeners.lua
index 897f4c1..ca8bbdf 100644
--- a/lua/diffview/scene/views/file_history/listeners.lua
+++ b/lua/diffview/scene/views/file_history/listeners.lua
@@ -59,6 +59,14 @@ return function(view)
     select_prev_entry = function()
       view:prev_item()
     end,
+    select_first_entry = function()
+      local entry = view.panel.entries[1]
+      if entry and #entry.files > 0 then view:set_file(entry.files[1]) end
+    end,
+    select_last_entry = function()
+      local entry = view.panel.entries[#view.panel.entries]
+      if entry and #entry.files > 0 then view:set_file(entry.files[#entry.files]) end
+    end,
     next_entry = function()
       view.panel:highlight_next_file()
     end,
```

yungthai:
Thanks for the  improvement suggestions.

I have simplified the diff file_panel listeners, as there is no need to support a motion count within that panel.

For the file history panel: 
There may be some benefit to navigate to the first or last file entry of +/- x number of commits, rather than just jump to the first and last file within the file history panel. I am ok with your simplified suggestion for file history too, if that's what you prefer.

sindrets:
> There may be some benefit to navigate to the first or last file entry of +/- x number of commits, rather than just jump to the first and last file within the file history panel.

No, not for an action named `select_{first|last}_file`. It should just do as its name implies, there's no reason for this action to use `v:count`. For file based jumping that respects `v:count` we already have `[count] select_{next|prev}_file`. It sounds like you want another action as well. Something like `[count] select_{next|prev}_commit`. That's fine, we can add that as well, but that should be a separate PR.

Also, please remove all unrelated formatting changes. It only adds noise to the diffs.

yungthai:
* Tidied up the commit and removed all unrelated formatting changes.
* Ensure the first | last file selected is highlighted within the diff file panel.

Will follow up with a separate PR for the `[count] select_{next|prev}_commit`.

sindrets (approved):
Thanks!

PR #557, opened 2025-02-19
fix: only replace path with env var if value defined

### LINKED ISSUE #556: [Bug] can't open files with path containing the `$` character
### Description

Files with a path containing a `$` character are not loaded in diffview. There are frontend routing frameworks that use the `$` character for file-based routing.

### Expected behavior

Files with a path containing a `$` character can be viewed in diffview like any other file.

### Actual behavior

Diffview seems to show the file under a path without the `$` character. For example a file at the path `$b/c` would become `b/c`. Diffview will show this as a file that has been removed because it doesn't exist on disk. `gf` will show the error

```
File does not exist on disk: 'b/c'
```

### Steps to reproduce

1. `mkdir -p a/\$b`
2. `cd a`
3. `git init`
4. `touch \$b/c`
5. `nvim \$b/c`
6. `:DiffviewOpen`
7. `gf`

### Health check

<details>
<summary>Output of <code>:checkhealth diffview</code></summary>

```
==============================================================================
diffview: require("diffview.health").check()

Checking plugin dependencies ~
- OK nvim-web-devicons installed.

Checking VCS tools ~
- The plugin requires at least one of the supported VCS tools to be valid.
- OK Git found.
- OK Git is up-to-date. (2.48.1)
- WARNING Configured `hg_cmd` is not executable: 'hg'
```

</details>

### Log info

<details>
<summary>Relevant info from <code>:DiffviewLog</code></summary>

```
[ERROR 2025-02-19 11:41:32.311 -0800] ...cal/share/nvim/lazy/diffview.nvim/lua/diffview/utils.lua:42: File does not exist on disk: 'b/c'
```

</details>

### Neovim version

```markdown
NVIM v0.10.4
Build type: Release
LuaJIT 2.1.1713484068
```

### Operating system and version

macOS Sonoma 14.4.1

### Minimal config

```Lua
-- #######################################
-- ### USAGE: nvim --clean -u mini.lua ###
-- #######################################

local root = vim.fn.stdpath("run") .. "/nvim/diffview.nvim"
local plugin_dir = root .. "/plugins"
vim.fn.mkdir(plugin_dir, "p")

for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

local plugins = {
  { "nvim-web-devicons", url = "https://github.com/nvim-tree/nvim-web-devicons.git" },
  { "diffview.nvim", url = "https://github.com/sindrets/diffview.nvim.git" },
  -- ##################################################################
  -- ### ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE ###
  -- ##################################################################
}

for _, spec in ipairs(plugins) do
  local install_path = plugin_dir .. "/" .. spec[1]
  if vim.fn.isdirectory(install_path) ~= 1 then
    if spec.url then
      print(string.format("Installing '%s'...", spec[1]))
      vim.fn.system({ "git", "clone", "--depth=1", spec.url, install_path })
    end
  end
  vim.opt.runtimepath:append(spec.path or install_path)
end

require("diffview").setup({
  -- ##############################################################################
  -- ### ADD DIFFVIEW.NVIM CONFIG THAT IS _NECESSARY_ FOR REPRODUCING THE ISSUE ###
  -- ##############################################################################
})

vim.opt.termguicolors = true
vim.cmd("colorscheme " .. (vim.fn.has("nvim-0.8") == 1 and "habamax" or "slate"))

-- ############################################################################
-- ### ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE ###
-- ############################################################################

print("Ready!")
```

bryankenote:
fixes https://github.com/sindrets/diffview.nvim/issues/556

The code in question is trying to replace an environment variable name in a file path with it's value. It is assuming that if a segment starts with `$` then it is an environment variable name. There are frontend frameworks with file-based routing that use `$` to indicate a value in a path (ie https://tanstack.com/router/latest/docs/framework/react/guide/file-based-routing). We need to strip the `$` off of the name to pass to `os_getenv` but If there isn't a value for the environment variable, we shouldn't remove the `$` from the path since this is a valid file and directory name.

### DIFF
--- a/lua/diffview/path.lua
+++ b/lua/diffview/path.lua
@@ -295,8 +295,8 @@ function PathLib:expand(path)
 
   for i = idx, #segments do
     local env_var = segments[i]:match("^%$(%S+)$")
-    if env_var then
-      segments[i] = uv.os_getenv(env_var) or env_var
+    if env_var and uv.os_getenv(env_var) ~= nil then
+      segments[i] = uv.os_getenv(env_var)
     end
   end
 
--- a/lua/diffview/tests/functional/pathlib_spec.lua
+++ b/lua/diffview/tests/functional/pathlib_spec.lua
@@ -248,7 +248,7 @@ describe("diffview.path", function()
       local pl = PathLib({ os = "unix" })
 
       eq("/lorem/ipsum/dolor/foo", pl:expand("~/foo"))
-      eq("foo/EXPANDED_FOO/EXPANDED_BAR/baz", pl:expand("foo/$VAR_FOO/$VAR_BAR/baz"))
+      eq("foo/EXPANDED_FOO/EXPANDED_BAR/$baz", pl:expand("foo/$VAR_FOO/$VAR_BAR/$baz"))
     end)
   end)
 


bryankenote:
If we don't want to modify the existing method this way, we could also either

1. pass a parameter to disable replacing segments with environment vars or
2. make a separate method and only call the environment var version in a context where you would expect this behavior.

I don't have any strong opinions on this. Just would like to fix this bug. Please let me know if you would like any changes or additional test cases.

raine:
The fix works for me. Thanks @bryankenote 

Hopefully this could be merged.

1orZero:
Bumping this as well. Hopefully this PR will be merged.

PR #566, opened 2025-03-26
feat(file_history): allow cycle within single commit on prev/next file

przepompownia:


przepompownia:
Still without tests and docs, so in draft.

Issue #532, opened 2024-09-11
Feature request: config for file status icons


simonmandlik:
Hi, thanks for the amazing plugin!

I couldn't find a way to change file status "icons" (`M`, `D`, `?`, ...). Would it be possible to add configuration for this?

PR #607, opened 2026-01-22
feat: add status_icons configuration option

antinomie8:
Lets the user configure the git status icons displayed in the file tree.
I doubt the `get_status_icon` function has anything to do in `hl.lua`, perhaps it should be moved into another file.
It lets the user define custom icons for git status letters as such, for instance (nerd fonts):
```lua
status_icons = {
	["M"] = "", -- modified
	["T"] = "T", -- type
	["A"] = "", -- added
	["D"] = "󱟃", -- deleted
	["R"] = "", -- renamed
	["C"] = "", -- copied
	["U"] = "", -- updated but unmerged
	["X"] = "X", --
	["B"] = "B", --
	["?"] = "", -- untracked
	["!"] = "", -- ignored
}
```


antinomie8:
closes https://github.com/sindrets/diffview.nvim/issues/532

